PR #26, opened 2015-02-05
clarify comments about MAXLOAD

cgtx:
The config file has a commented out value of 0.8 for `MAXLOAD`.  This implies that the default value is 0.8, which isn't the case.  We should modify the comments to indicate that the value defaults to the number of CPU cores.


lil-cain:
Sounds good to me - I wonder should the default be higher though? Ime, these logs are most useful when load is already high.

man-chung:
Closing, since I've changed the default value of MAX Value in pull request #31 and added the hashed out default value in racap.conf, which satisfies this requested change

PR #28, opened 2015-05-19
Change the way recaplog rotates files

marcosgildavid:
Hi,
The way recaplog is currently rotating files makes it really hard to use one-liners to parse files from previous days.
Can we change it to use a tgz of a folder, with the files for a particular day, instead of a single file with all the info concatenated? I don't thing this  will have much off an impact on the way it works but it will make troubleshooting issues a bit easier.
I ran into this problem while trying to troubleshoot a recurring issue on a server.

I've just changed a few lines of code to accomplish this.
I hope you find them useful and merge the pull request.

If you have any questions please let me now. 
Thanks!

### DIFF
--- a/recaplog
+++ b/recaplog
@@ -95,14 +95,18 @@ pack_old_logs() {
 		#
 		if [ $LOG_COMPRESS -eq 1 ]; then
 			print_msg DEBUG "do compress old log files"
-			cat ${BASEDIR}/${item}_${YESTERDAYDATE}-*.log | gzip > ${OUTPUTFILE}.gz
+			mkdir ${BASEDIR}/${item}_${YESTERDAYDATE}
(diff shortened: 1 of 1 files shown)


bhgraham:
looks good, someone with merge access will have to merge it.

thebwt:
Is this still relevant? The recaplog script has been altered a bit since then.

lil-cain:
@marcosgildavid has since left Rackspace, but looking at the code it looks like his original complaint is still valid - if we're interested in merging, I can make this apply to current master?

thebwt:
I'd first want to confirm that the problem still exists, but then yes. Can anyone confirm that this is still happening?

lil-cain:
https://github.com/rackerlabs/recap/blob/master/src/recaplog#L94-L104

Very much looks that way.

thebwt:
Cool beans, let's merge it and close this PR out

tonyskapunk:
Fixed in #72

PR #30, opened 2015-06-09
Change log parsing from .log* to _*.log

mrhillsman:
Using the current logfilename.log* returns an error:

[root@devsrv recap]# recaptool net 443
Information: Using /var/log/recap as recap log path
Information: Executing net 443 
ls: cannot access /var/log/recap/netstat.log*: No such file or directory
[root@devsrv recap]#

After modifying recaptool to parse as logfilename_*.log output succeeds:

[root@devsrv recap]# recaptool net 443
Information: Using /var/log/recap as recap log path
Information: Executing net 443 
2015-06-08_23:13:33	5
[root@devsrv recap]#

### DIFF
--- a/recaptool
+++ b/recaptool
@@ -8,7 +8,7 @@ procs_mem() {
 	PROCESS="$1"
 	TMP=`mktemp`
 	>$TMP
-	for i in `ls $RECAPPATH/ps.log*`; do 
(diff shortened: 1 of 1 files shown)


tonyskapunk:
This has been fixed in #51 and #52 thanks!

PR #37, opened 2015-11-02
Added a check to verify .mylogin.cnf file, if .my.cnf file is not used

fooltruth:
If servers are only configured to use login-paths current configuration fails. I've added checks to verify if .mylogin.cnf is used on the server.
Ref: https://dev.mysql.com/doc/mysql-utilities/1.6/en/mysql-utils-intro-connspec-mylogin.cnf.html

### DIFF
--- a/recap
+++ b/recap
@@ -87,6 +87,7 @@ DOTMYDOTCNF="/root/.my.cnf"
 USEFDISK="no"
 MYSQL_PROCESS_LIST="table"
 MIN_FREE_SPACE=0
+LOGINPATH="recap"
 
 ###########################################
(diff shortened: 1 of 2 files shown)


man-chung:
Per conversation, this will need to be rewritten to allow us to easily integrate this without rewriting large chunks of existing code.
Notes:
Hashed out in recap.conf with useful comments
If uncommented, then take priority over the default /root/.my.cnf - if statement around the 2 existing for loops.

tonyskapunk:
Checking this PR I see this is covered on 07eb9b93, the support for multiple `DOTMYDOTCNF` is now fixed with PR #68, because of that I'm closing this.

PR #38, opened 2015-11-09
log dirs to 0700

JayGoldberg:
recap captures data as root, but then places this data in a world-readable dir, and writes world-readable log files.

This leaks possibly confidential data, such as PID mappings to listening ports, probably others.

Perhaps new system group is a good idea "logreaders" or somesuch?

### DIFF
--- a/recap-installer
+++ b/recap-installer
@@ -41,9 +41,9 @@ echo
 mkdir -p -m0755 $SBINDIR
 mkdir -p -m0755 $DATADIR/doc/recap-$VERSION
 mkdir -p -m0755 $SYSCONFDIR/cron.d
-mkdir -p -m0755 $LOGDIR/recap
-mkdir -p -m0755 $LOGDIR/recap/backups
-mkdir -p -m0755 $LOGDIR/recap/snapshots
+mkdir -p -m0700 $LOGDIR/recap
+mkdir -p -m0700 $LOGDIR/recap/backups
+mkdir -p -m0700 $LOGDIR/recap/snapshots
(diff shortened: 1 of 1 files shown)


piersc:
Change looks good, and matches the ,spec file:

`%{__install} -dm0700 %{buildroot}%{_localstatedir}/log/recap`

 Verified on a test install. Thanks for the PR!!!

carlwgeorge:
The `recap` script sets the directory permissions to 750.

```
BASEDIR="/var/log/recap"
BACKUPDIR=${BASEDIR}/backups
SNAPSHOTSDIR=${BASEDIR}"/snapshots"
```
```
OUTPUTDIR_MODE="750"
```
```
create_output_dirs() {
	# create directories if they don't exist
	if [ ! -d $BASEDIR ]
	then
	  	mkdir -p $BASEDIR
	  	chmod $OUTPUTDIR_MODE $BASEDIR
	fi
	if [ ! -d $SNAPSHOTSDIR ]
	then
	  	mkdir -p $SNAPSHOTSDIR
	  	chmod $OUTPUTDIR_MODE $SNAPSHOTSDIR
	fi
	if [ ! -d $BACKUPDIR ]
	then
	  	mkdir -p $BACKUPDIR
	  	chmod $OUTPUTDIR_MODE $BACKUPDIR
	fi
}
```

I just finished the Makefile and used this same setting as the `recap` script.  I think either 700 or 750 is fine, but we need to make it the same everywhere:

* Makefile
* spec file
* debian rules file
* installer script (or we can just remove this file entirely so we are using the Makefile everywhere)

PR #41, opened 2016-01-20
Added 'sysstat' & 'iotop' to dependencies

seanorama:
Rackspace builds have sysstat by default, but other providers do not. (Former Racker here)


man-chung:
Thanks Sean!

PR #42, opened 2016-02-10
Updates to spec file as well as recaplog

buzzboy23:
spec: recaplog does not function from the cron file as it currently stands, as the RPM did not include installing it.
spec: added mailx and findutils as requirements (they provide binaries called from recaplog)
recaplog: changed mutt to mail, as most default system installs will have a 'mail' binary and consider 'mutt' as an extra add-on.

### DIFF
--- a/recap.spec
+++ b/recap.spec
@@ -8,7 +8,7 @@ Url: https://github.com/rackerlabs/%{name}
 Source0: https://github.com/rackerlabs/%{name}/archive/%{version}.tar.gz
 BuildArch: noarch
 %{?el5:BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)}
-Requires: sysstat, coreutils, procps, grep, gawk, bc, net-tools
+Requires: sysstat, coreutils, procps, grep, gawk, bc, net-tools, mailx, findutils
 
 
 %description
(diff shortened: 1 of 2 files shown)


recap.spec
```diff
@@ -24,6 +24,7 @@ optional reporting on Apache, MySQL, and network connections.
 %install
 %{?el5:%{__rm} -rf %{buildroot}}
 %{__install} -Dm0755 recap %{buildroot}%{_sbindir}/recap
+%{__install} -Dm0755 recaptool %{buildroot}%{_sbindir}/recaplog
```
carlwgeorge: This line needs `recaptool` changed to `recaplog`.

buzzboy23:
Thanks for the find!  Fixed, and tested my own rpmbuild this time around.  Hopefully this will pass muster!  (I deployed recap from a previous rpmbuild across about 100+ VMs, and root instantly started complaining about recaplog every day!  I'll be so happy when this one gets merged!)

carlwgeorge:
Looks like we partially addressed this (recaplog command) via the new Makefile (see #49).  Sorry for the trouble, but can you send a new pull request for just the mutt/mail switch to avoid the new conflicts?

tonyskapunk:
This is outdated, as #72 refactored `recaplog` now emails are not sent through the tool but through cronjob only.

PR #50, opened 2016-04-01
Update recap-installer

wesmattson:
Add if statement to ensure the recap.5 and recap.8 files exist before running gzip against them. Reasoning for this is because the first time you run recap-installer it works fine but if you run recap-installer again, it fails because it cannot find recap.5 and recap.8 since they've been gzipped to recap.5.gz and recap.8.gz.


tonyskapunk:
Thanks for the patch, the location of the files changed in a632d16c, hence this is not needed any longer.

Issue #39, opened 2015-12-10
change processor utilisation report; report inaccurate


kimvette:
Hi team! 

I'd like to point out that relying on 'ps' for processor utilisation can lead to incorrect conclusions and if used for alert generation, can raise tickets based on outdated data. Why? From the ps man page: 

> CPU usage is currently expressed as the percentage of time 
> spent running during the entire lifetime of a process.  
> This is not ideal, and it does not conform to the standards
> that ps otherwise conforms to.  CPU usage is unlikely to add 
> up to exactly 100%.

Looking at processsor utilisation via "ps" isn't worthless; it is useful for identifying long-term trends but consider this: you've had a server that has been up for a few months and had a very heavy load but it is a single server behind a load balancer, and now your budget allows you to add three more nodes (and actually make use of the load balancer!) so you add the servers and pool members but have not restarted the java process on the original node. You want to make sure that the load is evenly spread so you install recap for monitoring and have your ticketing system parse the logs to generate alerts where appropriate. Your ticketing system is now generating alerts for the original node because the CPU utilisation reported by PS is based on many months' worth of of 90%+ CPU utilisation and the average will take weeks to drop to your "warning" threshold. 

A better method would be to run something like "top -d 5 -n 2" (naturally additional switches will be required for full output), discarding the first report and capturing the second; this would give a more relevant metric to raise alerts on and eliminate false alerts resulting from a historical average that has been addressed but not yet reflected by ps -- or maybe there is another method which would be less cumbersome. 

But, the bottom line is relying on ps can result in false conclusions if you're trying to diagnose an issue, have applied some changes and are trying to figure out why it is still reporting high utilisation (when using ps) and generating alerts.

man-chung:
Hi,

Thank you for your feedback,

The design intention was not to alert for CPU utilisation - there are far more accurate, reliable, and timely tools/methods that fullfil that role (in real time, rather than waiting to parsing the recap output files).

Recap was designed to allow administrators a point in time view of the state of the server (likely, during the time of an incident/issue), much like if the admin was there to run the commands at the time. So as you mentioned, trend is more desirable from this point of view.

Beerey:
[...]

PR #48, opened 2016-02-29
Record point-in-time CPU usage, instead of average since boot (issue …

piersc:
…#39)

PR #57, opened 2016-05-21
Cpuusage

lil-cain:


lil-cain:
Rebased changes by @piersc in #48, and add version detection to work
around differences in pidstat output format between RHEL 6 & RHEL 7 as
pointed out by @rakevinwr. Closes #39 #48.

Beerey:
Assuming this needs to be RHEL5 compatible, the sorted_versions list you're generating which uses `sort --version-sort` won't work - it was introduced with RHEL6 sort.  (minor) but also missing an 'e' in the comment.

carlwgeorge:
This detects `version_strings` by using rpm.  That won't work if this is installed on a non-rpm distribution.

lil-cain:
@carlwgeorge If I add similar code to do this with dpkg, wil this suffice? Or do we aim to support non rpm/deb distros?

thebwt:
Sysstat -v should do this universally

thebwt:
Okay I take it back, this isn't easy to get the version of...

lil-cain:
Yeah - I tried looking for version flags before trying the rpm thing. I can add something similar to use dpkg to get version numbers, but if maybe we need to pull the data directly from /proc, which would be pretty horrible to have to do in shell…

thebwt:
``sar -V
sysstat version 10.2.0```

That works on my 14.04 server, how about on rhelish systems?

carlwgeorge:
This works (with sar or any of the other sysstat binaries):

```
$ pidstat -V 2>&1 | awk '{if(NR==1){print$NF}}'
9.0.4
```

piersc:
@lil-cain could you change the version detection in this PR to @carlwgeorge 's method?

lil-cain:
@piersc Done - does that look good to you? If so, I'll squash the commits together, and it should be good to merge?

piersc:
Testing now

piersc:
Looks good to me!

lil-cain:
should be ready to merge :-)

bhgraham:
There is a typo in the first line of the commit.  

> # print th top 10 processes (by cpu usage) to the resources file

lil-cain:
fixed - good spot

piersc:
Awesome, thanks, merged.

Issue #61, opened 2016-06-07
recap log.gz files don't rotate


Nohau:
I have noticed the recap gzip files don't rotate properly after looking into it the regex is wrong inside /usr/sbin/recaplog here is a example : 

find $BASEDIR -maxdepth 1 -regextype grep -regex "${BASEDIR}/[a-z]*_[0-9]{8}-[0-9]{6}.log.gz" -type f -mtime +$LOG_EXPIRY -exec echo rm {} \;  | tee -a $LOGFILE

The file name is the following that gets made when rotated : 

netstat_daily_20160602.log.gz

So as you can see there is no - and actually another _  . If we where rotating just the log file this would have worked but this is not the case : 

ps_20160607-012001.log

As you see it has the single _ and a single - between the dates .  

I created a temporary work around on my side by replacing the find regex

FROM:

-regex "${BASEDIR}/[a-z]*_[0-9]{8}-[0-9]{6}.log.gz" 

TO:

-regex "${BASEDIR}/[a-z]__[a-z]__[0-9]*.log.gz"

Now it actually finds the gzip files and it can rotate them out as per the configuration.

thebwt:
That's a pretty simple fix, why don't you open a pull request (make sure to mention "issue #61"), and then you can get the credit for the change.

tonyskapunk:
This should be addressed by #72

tonyskapunk:
Fixed in #72

PR #62, opened 2016-06-07
Issue #61 - Not rotating log.gz files

Nohau:
Fix for not rotating the gzip files once it gets rotated from .log to do .log.gz and it never cleans the gzip files after expire date due to invalid regex.  This is part of issue part of issue #61 that was reported.


thebwt:
looks good to me, i don't use the tool a lot though. we'll give people till Friday to review it.

tonyskapunk:
Fixed in #72

PR #66, opened 2016-10-07
Add recaptool man page.

### LINKED ISSUE #53: need manpage for recaptool
We have man pages for `recap` and `recaplog`, but not for `recaptool`.  Can someone familiar with that utility write one?

tonyskapunk:
Fix #53

### DIFF
--- a/src/recaptool.8
+++ b/src/recaptool.8
@@ -1,0 +1,126 @@
+.\"
+.\" This is free documentation; you can redistribute it and/or
+.\" modify it under the terms of the GNU General Public License as
+.\" published by the Free Software Foundation; either version 2 of
+.\" the License, or (at your option) any later version.
+.\"
+.\" The GNU General Public License's references to "object code"
+.\" and "executables" are to be interpreted as the output of any
+.\" document formatting or typesetting system, including
+.\" intermediate and printed output.
+.\"
+.\" This manual is distributed in the hope that it will be useful,
+.\" but WITHOUT ANY WARRANTY; without even the implied warranty of
+.\" MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+.\" GNU General Public License for more details.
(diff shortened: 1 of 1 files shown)


tonyskapunk:
Please hold on this one, there is an undocumented feature in the recaptool that is not in place on the man page, nor explicit in the help output of the tool, since I'd like to see PR #67 introduced first I can include both changes on this PR once that's merged.

tonyskapunk:
Alright, I've fixed the `man` page for `recaptool` including an undocumented feature, which is now clearly displayed on the tool itself via #69

tonyskapunk:
It has been a good time without any additional comment on this, I'm going to merge since is documentation.  It will need a change in the `spec` file and the `Makefile` to include `recaptool.8` as well, I'll take care of that in another PR.

carlwgeorge:
Before adding it to the spec file, a new version should be tagged so that way `recaptool.8` is in the tarball.

tonyskapunk:
Yeah, that's what I saw on the `spec` file, I was thinking in **not** bumping the version since there is no change in the code, just on the docs, if I don't bump the version I understand there will be no new package, what would be your suggestion in this case?

tonyskapunk:
@carlwgeorge created #71, might be useless, but I wanted to ensure we update all the pieces required for this new man page.

carlwgeorge:
Really comes down to two options.

1. Bump the version to get a new tarball with the man page included.
2. Copy the the man page to `util/packaging/rpm/` directory and list it as `Source1` in the spec file.

Option 1 is easy and simple.  We don't even have to rush it, and just wait to merge #71 when a new version is imminent.
Option 2 is messier, but certainly doable.  We would just have to remember to undo it when the next version happens.

tonyskapunk:
Yeah I rather wait, as many other changes are in coming.

Thanks!

PR #69, opened 2016-10-28
Documenting options of recaptool.

tonyskapunk:


tonyskapunk:
Haven't seen any comment on this, since this is just to make clear what are the functions already in recaptool I'm going to merge, this is documented in the man page(#66) already as well.

PR #71, opened 2016-12-16
Include recaptool man page on build.

tonyskapunk:
Per #66 I'm adding this PR, but since packaging relies on version tag this might be useless for building automatically the RPM, the purpose is to ensure the installer includes the new man page.


tonyskapunk:
Closing this as at this time we will not create a new RPM for this until (hopefully) we release `1.0`

PR #70, opened 2016-11-04
Silence STDOUT in recap.cron

StafDehat:
Should address https://github.com/rackerlabs/recap/issues/35

Issue #35, opened 2015-09-09
Is there any way to make it stop sending me an email daily?


mstevens:
Is there any way to make it stop sending me an email daily?

I just want the logs.

carlgeorge:
Yes, just unset the MAILTO variable in the config file.

https://github.com/rackerlabs/recap/blob/0.9.8/recap.conf#L16-L18
https://github.com/rackerlabs/recap/blob/0.9.8/recap#L718-L721

mstevens:
Tested this, I still get emailed every day by recaplog:

[DD] packing fdisk...
[DD] do compress old log files
cat: /var/log/recap/fdisk_20150909-*.log: No such file or directory
ls: cannot access /var/log/recap/fdisk_20150909-*.log: No such file or directory
[DD] packing mysql...
[DD] do compress old log files
[DD] deleting: /var/log/recap/mysql_20150909-000001.log
[DD] deleting: /var/log/recap/mysql_20150909-001001.log
[DD] deleting: /var/log/recap/mysql_20150909-002001.log
[DD] deleting: /var/log/recap/mysql_20150909-003001.log
[DD] deleting: /var/log/recap/mysql_20150909-004001.log
[DD] deleting: /var/log/recap/mysql_20150909-005002.log
[DD] deleting: /var/log/recap/mysql_20150909-010001.log
[DD] deleting: /var/log/recap/mysql_20150909-011002.log
[DD] deleting: /var/log/recap/mysql_20150909-012001.log
[DD] deleting: /var/log/recap/mysql_20150909-013001.log
[DD] deleting: /var/log/recap/mysql_20150909-014002.log
[DD] deleting: /var/log/recap/mysql_20150909-015001.log
[DD] deleting: /var/log/recap/mysql_20150909-020001.log
[DD] deleting: /var/log/recap/mysql_20150909-021001.log
[DD] deleting: /var/log/recap/mysql_20150909-022002.log
[DD] deleting: /var/log/recap/mysql_20150909-023002.log
[DD] deleting: /var/log/recap/mysql_20150909-024001.log
[DD] deleting: /var/log/recap/mysql_20150909-025001.log
.... and continues

On further inspection it's due to cron output rather than being an email directly generated by recap.

carlwgeorge:
Sorry @mstevens, somehow I never noticed your follow up comment.

[...]

Issue #59, opened 2016-05-26
recap is "touching" rs-sysmon package files - should recap obsolete rs-sysmon


tonyskapunk:
Noticed the [%post](https://github.com/rackerlabs/recap/blob/master/util/packaging/rpm/recap.spec#L50-L71) script is doing some changes if `rs-sysmon` files are found in the system where `recap` is being installed, based on those I submitted #58 but since there is no interaction between `recap` and `rs-sysmon` RPMs I think this is a good opportunity to define whether or not `recap` should `Obsoletes: rs-sysmon` ?

IMO seems like this is a good use case of the use of `Obsoletes` as `recap` is a [rename](https://github.com/rackerlabs/recap/wiki/All-about-recap-and-historical-information) of `rs-sysmon`

Comments?

man-chung:
Yes, we should Obsolete and remove the %post script. +1 from me

tonyskapunk:
PR #60 proposes `Conflicts` I suggest `Obsoletes` as mentioned in here https://github.com/rackerlabs/recap/pull/60#issuecomment-222166268 any comments, ideas or suggestions?

carlwgeorge:
[Best practice](https://fedoraproject.org/wiki/Packaging:Guidelines#Renaming.2FReplacing_Existing_Packages) is to put an upper limit on the obsoletes, just in case you ever need to revert it.  The last version of rs-sysmon I can find was 0.9.5-1.rs, so we should use:

```
Obsoletes: rs-sysmon < 0.9.5-2
```

Also, if we feel that recap is a "compatible enough replacement" for rs-sysmon, we should have it provide the rs-sysmon name.  That would allow users to continue to run `yum install rs-sysmon`, which would resolve to install recap instead.

```
Provides: rs-sysmon = %{version}-%{release}
```

bhgraham:
Definitely, that was always the goal to be the full replacement.  I am all for adding the provides and obsoletes.  

And if we are saying it is stable to be the replacement, we should version up to 1.0 as well.

tonyskapunk:
@carlwgeorge  @bhgraham thanks for the input, I'll make sure `Obsoletes` and `Provides` are included in #60.

Regarding bumping to `1.0` I'd wait as there are still some PR and issues reported, trying to catch up on those.

tonyskapunk:
Fix on #89

PR #72, opened 2017-01-06
Major refactor of recaplog.

tonyskapunk:
This is a _major_ refactor of recaplog.

- Added dry-run and verbose modes to its functionality.
- Now behaves more like recap as not producing output if everything goes fine.
- Does not produce output if no tty is found, this is quite useful when running from cron and avoid tons of mails, now all errors are documented in the log.
- There is a new log format including timestamps using RFC-3339 with seconds, e.g. "2016-12-23 01:00:07+00:00".
- The variables now honor the main config when defined.
- Logs instead of being concatenated are now put into a directory and can be compressed via `LOG_COMPRESS`
- Start and End banners are used in the logs to better understand the time it took to complete a task.

Fixes: 
- #61 , #62
- #35 , #70 
- #28 

A test of the functionality: https://gist.github.com/tonyskapunk/a9fa24d02cb6fa2ee3c8f0b106f0c747

TODO: Need to update `recaplog.8`


carlwgeorge:
Do any of these changes (such as the log format change) require manual steps to update from a prior version of recap?  If so, please consider creating a 1.0 milestone and linking this PR to it for clarity.

tonyskapunk:
@carlwgeorge I don't think there is any manual step involved, just changes to a log format and behaviour of the `recaplog` tool, in any case, because of what was discussed in #59 I'm creating 1.0 milestone to introduce these changes and have a new release.

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
+      prefix_msg="Would delete:"
+    fi
+    # Deleting files and empty directories if any.
+    # Using now posix-extended regex flavor to ease the matching of logs/gzips/tarballs
+    matched_files=$(
+      find ${BASEDIR} -maxdepth 2 -regextype posix-extended \
+        -type f -mtime +${LOG_EXPIRY} -regex \
+        "^${BASEDIR}/(.*/)?[a-z]+_([a-z]+_)?[0-9]{8}(-[0-9]{6})?\.log(\.tar)?(\.gz)?$"
+    )
+    if [[ -n ${matched_files} ]]; then
```
StafDehat: Quote this variable.  If it ends up as a null string, as this is written currently, you'll be running:
```
if [[ -n ]]; then
```
And that's a syntax error.  Anytime a variable might be null, it's critically important to quote it.

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
+      count_files=$( wc -l <<<"${matched_files}" )
+    fi
+    log INFO "${prefix_msg} ${count_files} log files."
+    if [[ ! ${DRYRUN} ]]; then
+      # Deleting old logs, compressed/tarballs and empty dirs.
+      for matched_file in ${matched_files}; do
+        rm -f ${matched_file}
+      done
```
StafDehat: Files with spaces in the path will break this for loop.  You'll need to either backup/reset/restore IFS, or iterate over "while read LINE" instead.  Also, quote the "matched_file" variable so those whitespace characters don't break the rm.

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
+    log INFO "Skipping old log file expiration..."
+  else
+    log INFO "Deleting log files older than ${LOG_EXPIRY} days..."
+    if [[ ${DRYRUN} ]]; then
+      prefix_msg="Would delete:"
+    fi
+    # Deleting files and empty directories if any.
+    # Using now posix-extended regex flavor to ease the matching of logs/gzips/tarballs
+    matched_files=$(
+      find ${BASEDIR} -maxdepth 2 -regextype posix-extended \
```
StafDehat: Quote BASEDIR here, in case it has spaces.

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
-		find $BASEDIR -maxdepth 1 -regextype grep -regex "${BASEDIR}/[a-z]*_[0-9]\{8\}-[0-9]\{6\}.log.gz" -type f -mtime +$LOG_EXPIRY -exec echo rm {} \;  | tee -a $LOGFILE
-		find $BASEDIR -maxdepth 1 -regextype grep -regex "${BASEDIR}/[a-z]*_[0-9]\{8\}-[0-9]\{6\}.log.gz" -type f -mtime +$LOG_EXPIRY -exec rm {} \;
-	fi
+  local prefix_msg='Deleting:'
+  local matched_files=''
+  local empty_dirs=''
+  local count_files=0
+  local count_dirs=0
+  if [[ ${LOG_EXPIRY} -eq 0 ]]; then
+    log INFO "Skipping old log file expiration..."
```
StafDehat: Add a "return" here and close your 'if' block.  In general, if your conditional block goes to the end of the function, just put a conditional 'return' at the start instead.  It minimizes indentation and makes the code easier to read.

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
+      count_files=$( wc -l <<<"${matched_files}" )
+    fi
+    log INFO "${prefix_msg} ${count_files} log files."
+    if [[ ! ${DRYRUN} ]]; then
+      # Deleting old logs, compressed/tarballs and empty dirs.
+      for matched_file in ${matched_files}; do
+        rm -f ${matched_file}
+      done
+      # Deleting empty dirs generated by recaplog when no compressing logs.
+      empty_dirs=$( find ${BASEDIR} -maxdepth 1 -regextype posix-extended \
```
StafDehat: Quote BASEDIR.

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
+                      -regex "^${BASEDIR}/[a-z]+_[a-z]+_[0-9]{8}$" -empty \
+                      -type d )
+      if [[ -n ${empty_dirs} ]]; then
```
StafDehat: Quote empty_dirs

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
+        count_dirs=$( wc -l <<<"${empty_dirs}" )
+      fi
+      log INFO "${prefix_msg} ${count_dirs} empty directories."
+      for empty_dir in ${empty_dirs}; do
+        rmdir "${empty_dir}"
+      done
```
StafDehat: Same thing here as the other for-loop.  I suggest using:
```
while read empty_dir; do
done <<<"${empty_dirs}"
```

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
+  fi
+  # Backing up only items defined in the config
+  for item in ${BACKUP_ITEMS}; do
+    YESTERDAY_ITEM_DIR="${BASEDIR}/${item}_daily_${YESTERDAYDATE}"
+    err=''
+    log INFO "Packing ${item}..."
+    log_count=$(
+      {
+        ls -1 ${BASEDIR}/${item}_${YESTERDAYDATE}-*.log | wc -l
+      } 2>&1; exit ${PIPESTATUS[0]}
```
StafDehat: I'm not a fan of this structure - particularly the use of 'exit'.  At a glance, it seems as if execution halts there, but we're really just capturing a variable.  Also, I've had issues in the past trying to parse the output of 'ls'.
What about this instead?
```
find "${BASEDIR}" -type f -name "${item_${YESTERDAYDATE}-\*.log" 2>&1 | wc -l
```
Then, instead of the next 'if' testing $?, just specifically test ${PIPESTATUS[0]}.

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
+    else
+      log INFO "Moving ${log_count} logs to: ${YESTERDAY_ITEM_DIR}"
+      err=$(
+        {
+          mkdir -p "${YESTERDAY_ITEM_DIR}";
+          mv ${BASEDIR}/${item}_${YESTERDAYDATE}-*.log \
+             ${YESTERDAY_ITEM_DIR};
+        } 2>&1
+      )
+      if [[ ! -z ${err} ]]; then
```
StafDehat: Quote err

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
+        log ERROR "An error ocurred while attempting to move logs ${err}"\
+                  ", skipping..."
+        continue
+      fi
+    fi
+    # Compressing logs
+    if [[ ${LOG_COMPRESS} -eq 1 ]]; then
```
StafDehat: I'd suggest testing for "-ne 1" and jumping straight to 'continue', to reduce the levels of nesting.

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
+      OUTPUTFILE="${YESTERDAY_ITEM_DIR}.log.tar.gz"
+      if [[ ${DRYRUN} ]]; then
+        log INFO "Would compress ${log_count} logs into: ${OUTPUTFILE}"
```
StafDehat: Could short-circuit here too, with a 'continue', to reduce nesting.

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
+      else
+        log INFO "Compressing ${log_count} logs into: ${OUTPUTFILE}"
+        err=$(
+          {
+            tar czf ${OUTPUTFILE} \
+                    ${YESTERDAY_ITEM_DIR}/*.log 2>/dev/null &&
+            touch -t ${YESTERDAYDATE}0000 ${OUTPUTFILE};
+          } 2>&1
+        )
+        if [[ ! -z ${err} ]]; then
```
StafDehat: Quote err

src/recaplog
```diff
@@ -20,145 +20,252 @@
...
+    # Moving logs to yesterday item's directory
+    if [[ ${DRYRUN} ]]; then
+      log INFO "Would move ${log_count} logs to: ${YESTERDAY_ITEM_DIR}"
+    else
+      log INFO "Moving ${log_count} logs to: ${YESTERDAY_ITEM_DIR}"
+      err=$(
+        {
+          mkdir -p "${YESTERDAY_ITEM_DIR}";
+          mv ${BASEDIR}/${item}_${YESTERDAYDATE}-*.log \
+             ${YESTERDAY_ITEM_DIR};
```
StafDehat: Quote the vars.

src/recaplog
StafDehat: In fact, just quote all the variables.  All instances of 'err', all conditional tests for null or non-null strings, all directory paths, etc.

tonyskapunk:
@StafDehat all he suggestions have been performed, including reducing nested loops/conditionals when possible, I tested the script with few datasets and looks good, here the output of the log produced on the last test that included two different scenarios: https://gist.github.com/tonyskapunk/5794226dcbda88744a8ffc16c0f6746f
 
Could you please take a second look? Thanks!

src/recaplog
```diff
@@ -20,145 +20,261 @@
...
+    if [[ "${DRYRUN}" ]]; then
+      log INFO "Would move ${log_count} logs to: ${YESTERDAY_ITEM_DIR}"
+    else
+      log INFO "Moving ${log_count} logs to: ${YESTERDAY_ITEM_DIR}"
+      err=$(
+        {
+          mkdir -p "${YESTERDAY_ITEM_DIR}";
+          mv "${BASEDIR}/${item}_${YESTERDAYDATE}"-*.log \
+             "${YESTERDAY_ITEM_DIR}";
+        } 2>&1
```
StafDehat: Consider stripping newlines from ${err}, here.

src/recaplog
```diff
@@ -20,145 +20,261 @@
...
+      continue
+    fi
+    # Compressing logs when enabled
+    log INFO "Compressing ${log_count} logs into: ${OUTPUTFILE}"
+    err=$(
+      {
+        tar czf "${OUTPUTFILE}" \
+                "${YESTERDAY_ITEM_DIR}"/*.log 2>/dev/null &&
+        touch -t ${YESTERDAYDATE}0000 "${OUTPUTFILE}";
+      } 2>&1
```
StafDehat: Consider stripping newlines from ${err}, here.

src/recaplog
```diff
@@ -20,145 +20,261 @@
...
+    )
+    if [[ ! -z "${err}" ]]; then
+      log ERROR "An error ocurred while attempting to compress logs ${err}, "\
+                "skipping..."
+      continue
+    fi
+    log INFO "Deleting ${log_count} logs."
+    rm_err=$( { rm -Rf "${YESTERDAY_ITEM_DIR}"; } 2>&1 >/dev/null )
```
StafDehat: Consider stripping newlines from ${rm_err}, here.

src/recaplog
```diff
@@ -20,145 +20,261 @@
...
+  local count_files=0
+  local count_dirs=0
+  if [[ "${LOG_EXPIRY}" -eq 0 ]]; then
+    log INFO "Skipping old log file expiration..."
+    return 0
+  fi
+  # Proceed with deletion of old logs
+  log INFO "Deleting log files older than ${LOG_EXPIRY} days..."
+  if [[ "${DRYRUN}" ]]; then
+    prefix_msg="Would delete:"
```
StafDehat: For the benefit of the reader, consider initializing prefix_msg to "" and adding an else here to set prefix_msg="Deleting:".

src/recaplog
```diff
@@ -20,145 +20,261 @@
...
+  matched_files=$(
+    find "${BASEDIR}" -maxdepth 2 -regextype posix-extended \
+      -type f -mtime +${LOG_EXPIRY} -regex \
+      "^${BASEDIR}/(.*/)?[a-z]+_([a-z]+_)?[0-9]{8}(-[0-9]{6})?\.log(\.tar)?(\.gz)?$"
+  )
+  if [[ -n "${matched_files}" ]]; then
+    count_files=$( wc -l <<<"${matched_files}" )
+  fi
+  log INFO "${prefix_msg} ${count_files} log files."
+  if [[ ! "${DRYRUN}" ]]; then
```
StafDehat: Could short-circuit here to reduce nesting.  In fact, you could scrap "prefix_msg" altogether, and right here do a single conditional to report and maybe bail:
```
if [[ "${DRYRUN}" ]]; then
  log INFO "Would delete: ${count_files} log files."
  return
else
  log INFO "Deleting ${count_files} log files."
fi

tonyskapunk:
Followed suggestions from last feedback, see all changes on 037398a, I modified previous commit to have a better description.  I tested the functionality and the new implementation of the lock in here https://gist.github.com/tonyskapunk/b222a85b70b6f0a36b7861a2ae0d6e15.

StafDehat:
I like it. :+1:

tonyskapunk:
Finally Merging :D 

This fixes #61 , #62, #35 , #70 and #28 

Initial Functionality test: https://gist.github.com/tonyskapunk/a9fa24d02cb6fa2ee3c8f0b106f0c747
Latest Functionality test: https://gist.github.com/tonyskapunk/b222a85b70b6f0a36b7861a2ae0d6e15.

PR #75, opened 2017-01-20
Updating cronjob info.

### LINKED ISSUE #74: Execution via Cron is enabled by default
The readme clearly states that: "By default the cron execution of the program is not enabled."
Yet in the file https://github.com/rackerlabs/recap/blob/master/src/recap.cron which is copied to /etc/cron.d during installation, the following line is not commented:
*/10 * * * * root /usr/sbin/recap

tonyskapunk:
Fix #74

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,10 @@ The information captured will be found in log files in the `/var/log/recap/` dir
 
 ## Configuration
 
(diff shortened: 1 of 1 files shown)


tonyskapunk:
I'm merging since this change does not affect behaviour of the app, instead it is a change in the documentation to reflect the actual way the tool behaves.

PR #78, opened 2017-01-27
Better support for BASEDIR, in order to allow overriding the variable.

### LINKED ISSUE #63: Log output paths are set using $BASEDIR variable before custom .conf has a chance to set it
Therefore you end up with a hybrid variable set, such as:

BASEDIR -> /home/jay/log
PS_LOG -> /var/log/recap/ps-20160607.log

tonyskapunk:
At the same time multiple changes occurred:
- Standardization of variables.
- Removal of tabs for spaces on the functions modified.
- Reduction of declaration of variables to be more dynamic.
- Mail function reduced based on BACKUP_ITEMS.
- Created a more generic function to run_<ITEM> to call print_<ITEM> functions

Fix #63

### DIFF
--- a/src/recap
+++ b/src/recap
@@ -37,7 +37,7 @@ DATE=`date +%Y-%m-%d_%H:%M:%S`
 ROTATELOGS="yes"
 SNAPSHOT="no"
 BACKUP="no"
-lockdir=/var/lock/recap.lock
+LOCKFILE="/var/lock/recap.lock"
 BACKUP_ITEMS="fdisk mysql netstat ps pstree resources"
 MAXLOAD="1000"
 
@@ -92,17 +92,22 @@ OUTPUTDIR_MODE="750"
 ######### BEGIN FUNCTION DEFINITIONS #########
 ##############################################
 
-# Create a Lock so that recap does not try to run over itself, and create errors:
+
+# Cleanup function to remove lock
+cleanup() {
+  rm -f "${LOCKFILE}"
+  exit 2
+}
+
+# Create a Lock so that recap does not try to run over itself.
 recaplock() {
-	if mkdir "$lockdir"
-	then
-		# Remove lockdir when the script finishes, or when it receives a signal
-		trap 'rm -rf "$lockdir"' 0    # remove directory when script finishes
-		trap "exit 2" 1 2 3 15        # terminate script when receiving signal
-	else
-		echo "Cannot acquire lock, giving up on $lockdir"
-		exit 0
-fi
+  (set -C; echo "$$" > "${LOCKFILE}") 2>/dev/null
+  if [ $? -ne 0 ]; then
+    echo "ERROR $( basename $0 ) ($$): Lock File exists - exiting"
+    exit 1
+  else
+    trap 'cleanup' 1 2 15 17 19 23 EXIT
+  fi
 }
 
 # ensure our output directories exist before we start creating files
@@ -248,24 +253,15 @@ print_netstat_sum() {
 # print the output of "mysqladmin status" to the mysql file
 print_mysql() {
   local LOGFILE="$1"
-  # We only support a single DOTMYDOTCNF for MySQL when running Plesk
   local PLESK_FILE="/etc/psa/.psa.shadow"
   if [[ -r "${PLESK_FILE}" ]]; then
     echo "MySQL status" >> "${LOGFILE}"
     mysqladmin -uadmin -p$(cat "${PLESK_FILE}") status >> "${LOGFILE}"
-    print_blankline "${LOGFILE}"
   else
-    for cnf in $( sed -e 's/,/ /g' <<<"${DOTMYDOTCNF}" ); do
-      if [[ -r "${cnf}" ]]; then
-        echo "MySQL ${cnf} status" >> "${LOGFILE}"
-        mysqladmin --defaults-extra-file="${cnf}" status >> "${LOGFILE}"
-        print_blankline "${LOGFILE}"
-      else
-        echo "MySQL ${cnf} for status does not exist. Please check DOTMYDOTCNF in /etc/recap" >> "${LOGFILE}"
-        print_blankline "${LOGFILE}"
-      fi
-    done
+    echo "MySQL ${DOTMYDOTCNF} status" >> "${LOGFILE}"
+    mysqladmin --defaults-extra-file="${DOTMYDOTCNF}" status >> "${LOGFILE}"
   fi
+  print_blankline "${LOGFILE}"
 }
 
 # print the non-truncated innodb status to the mysql file
@@ -278,9 +274,11 @@ print_mysql_innodb_status() {
 
   echo "MySQL InnoDB status"  >> "${LOGFILE}"
   unset MYVALS PID_FILE TMPDIR
-  MYVALS=( $( IFS=$'\t' mysql -Bse "SHOW VARIABLES LIKE 'pid_file'; \
+  MYVALS=( $( IFS=$'\t' mysql --defaults-extra-file="${DOTMYDOTCNF}" \
+                              -Bse "SHOW VARIABLES LIKE 'pid_file'; \
                                     SHOW VARIABLES LIKE 'tmpdir'; \
-                                    SHOW ENGINE INNODB STATUS;" 2>/dev/null \
+                                    SHOW ENGINE INNODB STATUS;" \
+                              2>/dev/null \
                 | head -n 2 ) )
   PID_FILE="${MYVALS[1]%/}"
   TMPDIR="${MYVALS[3]%/}"
@@ -303,34 +301,28 @@ print_mysql_innodb_status() {
 # print the output of "mysqladmin processlist" to the mysql file
 print_mysql_procs() {
   local LOGFILE="$1"
-  # We only support a single DOTMYDOTCNF for MySQL when running Plesk
   local PLESK_FILE="/etc/psa/.psa.shadow"
   if [[ -r "${PLESK_FILE}" ]]; then
     echo "MySQL processes" >> "${LOGFILE}"
     if [[ ${MYSQL_PROCESS_LIST} == "table" ]]; then
-      mysqladmin -uadmin -p$( cat "${PLESK_FILE}" ) -v processlist >> "${LOGFILE}"
+      mysqladmin -uadmin -p$( cat "${PLESK_FILE}" ) -v processlist \
+        >> "${LOGFILE}"
     fi
     if [[ ${MYSQL_PROCESS_LIST} == "vertical" ]]; then
-      mysql -uadmin -p$( cat "${PLESK_FILE}" ) -e "show full processlist\G" >> "${LOGFILE}"
+      mysql -uadmin -p$( cat "${PLESK_FILE}" ) -e "show full processlist\G" \
+        >> "${LOGFILE}"
     fi
-    print_blankline "${LOGFILE}"
   else
-    for cnf in $( sed -e 's/,/ /g' <<<"${DOTMYDOTCNF}" ); do
-      if [[ -r "${cnf}" ]]; then
-        echo "MySQL ${cnf} processes" >> "${LOGFILE}"
-        if [[ ${MYSQL_PROCESS_LIST} == "table" ]]; then
-          mysqladmin --defaults-extra-file="${cnf}" -v processlist >> "${LOGFILE}"
-        elif [[ ${MYSQL_PROCESS_LIST} == "vertical" ]]; then
-          # This doesnt seem to work right
-          mysqladmin --defaults-extra-file="${cnf}" -e "show full processlist\G"
-        fi
-        print_blankline "${LOGFILE}"
-      else
-        echo "MySQL ${cnf} for processes does not exist. Please check DOTMYDOTCNF in /etc/recap" >> "${LOGFILE}"
-        print_blankline "${LOGFILE}"
-      fi
-    done
+    echo "MySQL ${DOTMYDOTCNF} processes" >> "${LOGFILE}"
+    if [[ ${MYSQL_PROCESS_LIST} == "table" ]]; then
+      mysqladmin --defaults-extra-file="${DOTMYDOTCNF}" -v processlist \
+        >> "${LOGFILE}"
+    elif [[ ${MYSQL_PROCESS_LIST} == "vertical" ]]; then
+      mysql --defaults-extra-file="${DOTMYDOTCNF}" \
+            -e "show full processlist\G" >> "${LOGFILE}"
+    fi
   fi
+  print_blankline "${LOGFILE}"
 }
 
 # print the top 10 processes (by cpu usage) to the defined file
@@ -468,7 +460,7 @@ run_resources_report() {
   if [[ "${USESARQ,,}" == "yes" ]]; then
     # send sar -q output to output file
     print_blankline "${ITEM_FILE}"
-    print_sar "${ITEM_FILE}" "q" 
+    print_sar "${ITEM_FILE}" "q"
   fi
 
   # check to see if apachectl fullstatus should be run
@@ -528,6 +520,13 @@ run_mysql_report() {
 
   create_output_file "${ITEM_FILE}"
   check_output_file "${ITEM_FILE}"
+  # Don't run reports if can't read the config file
+  if [[ ! -r "${DOTMYDOTCNF}" ]]; then
+    echo "Error when attempting to read: ${DOTMYDOTCNF}, please check the" \
+         "DOTMYDOTCNF setting in the configuration file" >> "${ITEM_FILE}"
+    return
+  fi
+
   print_mysql "${ITEM_FILE}"
 
   # check to see if the optional mysql process list should be generated
--- a/src/recap
+++ b/src/recap
@@ -397,8 +397,9 @@ send_mail() {
   # echo report files into temporary file for mailing
   for item in ${BACKUP_ITEMS}; do
     ITEM_FILE="${BASEDIR}/${item}_${LOG_SUFFIX}.log"
-    ITEM_USE=$( eval "echo \${USE${item^^}}" )
-    if [[ -r "${ITEM_FILE}" && "${ITEM_USE,,}" == "yes" ]]; then
+    # Build the variable to use and then use indirect ref to obtain its value
+    ITEM_USE="USE${item^^}"
+    if [[ -r "${ITEM_FILE}" && "${!ITEM_USE,,}" == "yes" ]]; then
       echo "-----BEGIN ${item^^} REPORT-----" >> "${REPORT}"
       cat "${ITEM_FILE}" >> "${REPORT}"
       echo "-----END ${item^^} REPORT-----" >> "${REPORT}"
--- a/src/recap
+++ b/src/recap
@@ -585,7 +585,7 @@ else
 fi
 
 # Quit if load is higher than MAXLOAD
-LOAD=$( grep -Po '^[\d.]+' /proc/loadavg )
+LOAD=$( grep -Po '^[\d\.]+' /proc/loadavg )
 QUIT=$( bc -l <<< "${LOAD}>${MAXLOAD}" )
 if [[ ${QUIT} -eq 1 ]]; then
   echo "QUIT (LOAD:"${LOAD}", MAXLOAD:"${MAXLOAD}")"
--- a/src/recap
+++ b/src/recap
@@ -114,7 +114,7 @@ create_output_dirs() {
   OUTPUT_DIRS+=( "${BACKUPDIR}" )
   OUTPUT_DIRS+=( "${SNAPSHOTSDIR}" )
   for OUTPUT_DIR in ${OUTPUT_DIRS[@]}; do
-    if [[ ! -d ${OUTPUT_DIR} ]]; then
+    if [[ ! -d "${OUTPUT_DIR}" ]]; then
       mkdir -p "${OUTPUT_DIR}"
       chmod ${OUTPUTDIR_MODE} "${OUTPUT_DIR}"
     fi
@@ -134,7 +134,7 @@ print_usage() {
 
 # create output file
 create_output_file() {
-  local OUTPUT_FILE=$1
+  local OUTPUT_FILE="$1"
   if [[ -d "${OUTPUT_FILE}" ]]; then
     echo "Target file already exists: ${OUTPUT_FILE}"
     exit
@@ -146,8 +146,8 @@ create_output_file() {
 
 # check to see if the output directory exists
 check_output_file() {
-  local OUTPUT_FILE=$1
-  if [[ ! -f "${OUTPUT_FILE}" ]]; then
+  local OUTPUT_FILE="$1"
+  if [[ ! -r "${OUTPUT_FILE}" ]]; then
     echo "The output file does not exist: ${OUTPUT_FILE}"
     exit
   fi
@@ -155,100 +155,99 @@ check_output_file() {
 
 # print output of "ps auxfww" to the specified file
 print_ps() {
-  local LOGFILE=$1
+  local LOGFILE="$1"
   ps ${OPTS_PS} >> "${LOGFILE}"
 }
 
 # print output of "pstree" to specified file
 print_pstree() {
-  local LOGFILE=$1
+  local LOGFILE="$1"
   pstree ${OPTS_PSTREE} >> "${LOGFILE}"
 }
 
 # print a blank line to the specified file
 print_blankline() {
-  local LOGFILE=$1
+  local LOGFILE="$1"
   echo " " >> "${LOGFILE}"
 }
 
 # print the output of "uptime" to the specified file
 print_uptime() {
-  local LOGFILE=$1
+  local LOGFILE="$1"
   echo "UPTIME report" >> "${LOGFILE}"
   uptime >> "${LOGFILE}"
 }
 
 # print the output of "free" to the specified file
 print_free() {
-  local LOGFILE=$1
+  local LOGFILE="$1"
   echo "FREE report" >> "${LOGFILE}"
   free ${OPTS_FREE} >> "${LOGFILE}"
 }
 
 # print the output of "vmstat" to the specified file
 print_vmstat() {
-  local LOGFILE=$1
+  local LOGFILE="$1"
   echo "VMSTAT report" >> "${LOGFILE}"
   vmstat ${OPTS_VMSTAT} >> "${LOGFILE}"
 }
 
 # print the output of "iostat" to the specified file
 print_iostat() {
-  local LOGFILE=$1
+  local LOGFILE="$1"
   echo "IOSTAT report" >> "${LOGFILE}"
   iostat ${OPTS_IOSTAT} >> "${LOGFILE}"
 }
 
 # print the output of "iotop" to the specified file
 print_iotop() {
-  local LOGFILE=$1
-  which iotop >/dev/null 2>&1 || return 1
+  local LOGFILE="$1"
   echo "IOTOP report" >> "${LOGFILE}"
   iotop ${OPTS_IOTOP} >> "${LOGFILE}"
 }
 
 # print the output of sar to the specified file
 print_sar() {
-  local OPTION=$1
-  local LOGFILE=$2
+  local LOGFILE="$1"
+  local OPTION="$2"
   local FLAGS=''
   # check to see if we're going to use any parameters for sar
-  if [[ ${OPTION} == "r" ]]; then
-    FLAGS=" -r"
-  elif [[ ${OPTION} == "q" ]]; then
-    FLAGS=" -q"
+  if [[ "${OPTION}" == "r" ]]; then
+    FLAGS="-r"
+  elif [[ "${OPTION}" == "q" ]]; then
+    FLAGS="-q"
   else
     FLAGS=""
   fi
 
   echo "SAR${FLAGS} report" >> "${LOGFILE}"
-  sar${FLAGS} >> "${LOGFILE}"
+  sar "${FLAGS}" >> "${LOGFILE}"
 }
 
 # print the output of "apachectl fullstatus" to the specified file
 print_httpd_fullstatus() {
-  local LOGFILE=$1
+  local LOGFILE="$1"
   echo "Apache Status report" >> "${LOGFILE}"
   apachectl fullstatus >> "${LOGFILE}"
 }
 
 # print the output of "netstat -ntulpae" to the specified file
 print_netstat() {
-  local LOGFILE=$1
+  local LOGFILE="$1"
   echo "Network connections" >> "${LOGFILE}"
   netstat ${OPTS_NETSTAT} >> "${LOGFILE}"
 }
 
 # print the output of "netstat -s" to the specified file
 print_netstat_sum() {
-  local LOGFILE=$1
+  local LOGFILE="$1"
   echo "Network traffic summary" >> "${LOGFILE}"
   netstat ${OPTS_NETSTAT_SUM} >> "${LOGFILE}"
 }
 
 # print the output of "mysqladmin status" to the mysql file
 print_mysql() {
-  local LOGFILE=$1
+  local LOGFILE="$1"
   # We only support a single DOTMYDOTCNF for MySQL when running Plesk
   local PLESK_FILE="/etc/psa/.psa.shadow"
   if [[ -r "${PLESK_FILE}" ]]; then
@@ -271,7 +270,7 @@ print_mysql() {
 
 # print the non-truncated innodb status to the mysql file
 print_mysql_innodb_status() {
-  local LOGFILE=$1
+  local LOGFILE="$1"
   # First, we establish which tmpdir and pid_file are in use, strip any
   # trailing slash and populate the file with current information.
   # We throw away the output of the "show engine innodb status" command as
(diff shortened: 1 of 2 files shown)


src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+  pidstat -l 2 2 | grep -v '%system' |
+    egrep ^Average: | sort -nr -k ${pidstat_cpufield} |
+    head -11 >> "${LOGFILE}"
+}
+
+# print the top 10 processes (by memory usage) to the defined file
 print_top_10_mem() {
-	echo "Top 10 memory using processes" >> $RESOURCES_FILE
-	ps auxww --sort=-rss|head -11 >> $RESOURCES_FILE
+  local LOGFILE=$1
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+  echo "Top 10 memory using processes" >> "${LOGFILE}"
+  ps auxww --sort=-rss | head -11 >> "${LOGFILE}"
 }
 
-# print the disk utilization to the resources file
+# print the disk utilization to the defined file
 print_df() {
-	echo "Disk Utilization" >> $RESOURCES_FILE
-	df $OPTS_DF >> $RESOURCES_FILE
+  local LOGFILE=$1
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+  echo "Disk Utilization" >> "${LOGFILE}"
+  df ${OPTS_DF} >> "${LOGFILE}"
 }
 
 # print the disk partitions to the fdisk file
 print_fdisk() {
-	echo "Disk partitions" >> $FDISK_FILE
-	fdisk $OPTS_FDISK >> $FDISK_FILE
+  local LOGFILE=$1
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+  echo "Disk partitions" >> "${LOGFILE}"
+  fdisk ${OPTS_FDISK} >> "${LOGFILE}"
 }
 
-# print the slabinfo command to the resources file
+# print the slabinfo command to the defined file
 print_slabinfo() {
-	echo "Slab Information" >> $RESOURCES_FILE
-	printf "name\tactive\tnum_obj\tobj_size\thuh\n" >> $RESOURCES_FILE; tail -n+2 /proc/slabinfo | awk 'size=$3*$4 {print $1"\t"$2"\t"$3"\t"$4"\t"size/1048576}' | sort -k5gr >> $RESOURCES_FILE
+  local LOGFILE=$1
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+  # create a temp directory
+  local TEMPDIR=$( mktemp -d "/tmp/rs.XXXXXXXXXX" )
+  local REPORT="${TEMPDIR}/report.log"
+  local ITEM_FILE=""
+  local ITEM_USE=""
+
+  # echo report files into temporary file for mailing
+  for item in ${BACKUP_ITEMS}; do
+    ITEM_FILE="${BASEDIR}/${item}_${LOG_SUFFIX}.log"
+    ITEM_USE=$( eval "echo \${USE${item^^}}" )
```
StafDehat: Don't need eval here.

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+    if [[ -f "${ITEM_FILE}" && ${ITEM_USE,,} == "yes" ]]; then
```
StafDehat: Quote ITEM_USE

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+  local item="$1"
+  local ITEM_FILE="${BASEDIR}/${item}_${LOG_SUFFIX}.log"
 
-# manage fdisk report
-run_fdisk_report() {
-	if [ $SNAPSHOT = "yes" ]
-	then
-		FDISK_FILE="$SNAPSHOTSDIR/fdisk.log_snapshot_$DATE";
-	fi
+  if [[ ${SNAPSHOT,,} == "yes" ]]; then
```
StafDehat: Quote SNAPSHOT

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+    ITEM_FILE="${SNAPSHOTSDIR}/${item}.log_snapshot_${DATE}"
```
StafDehat: SNAPSHOTSDIR only existed inside the create_output_dirs function.  It's no longer a global variable, after your changes.

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
-	fi
-
-	print_blankline $RESOURCES_FILE
-	print_top_10_cpu
-	print_blankline $RESOURCES_FILE
-	print_top_10_mem
+  local item="resources"
+  local ITEM_FILE="${BASEDIR}/${item}_${LOG_SUFFIX}.log"
+
+  if [[ ${SNAPSHOT,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+  print_blankline "${ITEM_FILE}"
+  print_vmstat "${ITEM_FILE}"
+  print_blankline "${ITEM_FILE}"
+  print_iostat "${ITEM_FILE}"
+  print_blankline "${ITEM_FILE}"
+  print_iotop "${ITEM_FILE}"
+  print_blankline "${ITEM_FILE}"
+
+  # check to see if sar should be run
+  if [[ ${USESAR,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+    print_blankline "${ITEM_FILE}"
+    print_sar "${ITEM_FILE}"
+  fi
+
+  # check to see if sar -r should be run
+  if [[ ${USESARR,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+    # send sar -r output to output file
+    print_blankline "${ITEM_FILE}"
+    print_sar "r" "${ITEM_FILE}"
+  fi
+
+  # check to see if sar -q should be run
+  if [[ ${USESARQ,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+    # send sar -q output to output file
+    print_blankline "${ITEM_FILE}"
+    print_sar "q" "${ITEM_FILE}"
+  fi
+
+  # check to see if apachectl fullstatus should be run
+  type -p 'apachectl' > /dev/null
+  if [[ $? -eq 0 && ${USEFULLSTATUS,,} == "yes" ]]; then
```
StafDehat: Quotes around USEFULLSTATUS

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+    # send apachectl fullstatus output to output file
+    print_blankline "${ITEM_FILE}"
+    print_httpd_fullstatus
+  fi
+
+  if [[ ${USEDF,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+    # send df -h output to output file
+    print_blankline "${ITEM_FILE}"
+    print_df "${ITEM_FILE}"
+  fi
+
+  if [[ ${USESLAB,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
-		if [ $USENETSTATSUM = "yes" -o $USENETSTATSUM = "YES" ]
-		then
-			print_blankline $NETSTAT_FILE
-			print_netstat_sum
-		fi
-	fi
+  local item="netstat"
+  local ITEM_FILE="${BASEDIR}/${item}_${LOG_SUFFIX}.log"
+
+  if [[ ${SNAPSHOT} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+    ITEM_FILE="${SNAPSHOTSDIR}/${item}.log_snapshot_${DATE}"
```
StafDehat: SNAPSHOTSDIR is undefined.

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+  fi
+
+  create_output_file "${ITEM_FILE}"
+  check_output_file "${ITEM_FILE}"
+  print_netstat "${ITEM_FILE}"
+
+  # check to see if optional netstat summary report should be run
+  if [[ ! -z ${USENETSTATSUM} ]]; then
+    if [[ ${USENETSTATSUM,,} == "yes" ]]; then
```
StafDehat: Quotes here, and you can completely remove the test for '! -z'

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
-	if [ $USEINNODB = "yes" -o $USEINNODB = "YES" ]
-	then
-		# send df -h output to output file
-		print_blankline $MYSQL_FILE
-		print_mysql_innodb_status
-	fi
+  local item="mysql"
+  local ITEM_FILE="${BASEDIR}/${item}_${LOG_SUFFIX}.log"
+
+  if [[ ${SNAPSHOT} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+    ITEM_FILE="${SNAPSHOTSDIR}/${item}.log_snapshot_${DATE}"
+  fi
+
+  create_output_file "${ITEM_FILE}"
+  check_output_file "${ITEM_FILE}"
+  print_mysql "${ITEM_FILE}"
+
+  # check to see if the optional mysql process list should be generated
+  if [[ ! -z ${USEMYSQLPROCESSLIST} ]]; then
+    if [[ ${USEMYSQLPROCESSLIST,,} == "yes" ]]; then
```
StafDehat: Quotes here, and you can completely remove the test for '! -z'

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+      print_blankline "${ITEM_FILE}"
+      print_mysql_procs "${ITEM_FILE}"
+    fi
+  fi
+  if [[ ${USEINNODB,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
-	for f in $BACKUP_ITEMS; do
-		ls -1t ${BASEDIR}/${f}_*.log | \
-		grep ".*_[0-9]\{8\}-[0-9]\{6\}.log" | \
-		head -1 | \
-		xargs -I {} cp {} $BACKUPDIR
-	done
+  for f in ${BACKUP_ITEMS}; do
+    ls -1t ${BASEDIR}/${f}_*.log | \
+    grep -m1 ".*_[0-9]\{8\}-[0-9]\{6\}.log" | \
+    xargs -I {} cp {} ${BACKUPDIR}
```
StafDehat: Seems needlessly complex.  Why not skip the grep altogether and "head -n 1"?

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
-        	FREE_SPACE=${FREE_SPACE%M}
-        	if [[ $MIN_FREE_SPACE -ge $FREE_SPACE ]]
-			then
-        		wall "Unable to run recap due to not enough disk space"
-        	fi
-        fi
+  if [[ ${MIN_FREE_SPACE} -eq 0 ]]; then
+    return 0
+  else
+    FREE_SPACE=$( df -BM ${BASEDIR} | grep -v 1M-blocks | awk '{print $4}' )
```
StafDehat: Quotes on BASEDIR here.  Also, you can avoid piping to awk by specifying the column you want:
```
df -BM --output=avail "${BASEDIR}" | tail -n 1
```

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+    FREE_SPACE=${FREE_SPACE%M}
+    if [[ ${MIN_FREE_SPACE} -ge ${FREE_SPACE} ]]; then
```
StafDehat: Gonna need quotes here, because "check_disk_space" runs prior to "create_output_dirs", so it's possible you just ran 'df' on a directory that doesn't exist, which would cause ${FREE_SPACE} to be null.

src/recap
```diff
@@ -643,87 +579,83 @@ check_disk_space() {
...
 
 # Quit if load is higher than MAXLOAD
-LOAD=$(uptime | awk -F'load average:' '{ print $2 }' | cut -d" " -f2 | cut -d"," -f1)
-QUIT=$(echo $LOAD'>'$MAXLOAD | bc -l)
-if [ $QUIT -eq 1 ]
-then
-    echo "QUIT (LOAD:"$LOAD", MAXLOAD:"$MAXLOAD")"
-    exit 1
+LOAD=$( uptime | awk -F'load average:' '{ print $2 }' |
+          cut -d" " -f2 | cut -d"," -f1 )
```
StafDehat: Recommend using this instead: grep -Po '^[\d\.]+' /proc/loadavg

src/recap
```diff
@@ -643,87 +579,83 @@ check_disk_space() {
...
+QUIT=$( echo ${LOAD}'>'${MAXLOAD} | bc -l )
```
StafDehat: Don't echo to pipes.  We know herestrings now: bc -l <<<"${LOAD}>${MAXLOAD}"

src/recap
```diff
@@ -643,87 +579,83 @@ check_disk_space() {
...
-then
-	create_backup
-	exit 0
-fi
-
-# proceed to report generation
-
 # grab the server's host name
-HOSTNAME=`hostname`
+HOSTNAME=$( hostname )
```
StafDehat: Quotes

src/recap
```diff
@@ -643,87 +579,83 @@ check_disk_space() {
...
 
 # create output directory if it is not already present
 create_output_dirs
 
+# take a backup?
+if [[ ${BACKUP,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -734,53 +666,44 @@ create_output_dirs
 # -Update the documentation to reflect the new report				  #
 ###################################################################################
 
+recaplock
 # run the ps report
-if [ $USEPS = "yes" -o $USEPS = "YES" ]
-then
-	# run the ps report
-	run_ps_report
+if [[ ${USEPS,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -734,53 +666,44 @@ create_output_dirs
...
+  run_item_report "ps"
 fi
 
 # run the resources report
-if [ $USERESOURCES = "yes" -o $USERESOURCES = "YES" ]
-then
-	# run the resources report
-	run_resources_report
+#TODO: standardize the run_resources_report
+if [[ ${USERESOURCES,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -734,53 +666,44 @@ create_output_dirs
...
+  run_resources_report
 fi
 
-# check to see if pstree output should be gathered
-if [ $USEPSTREE = "yes" -o $USEPSTREE = "YES" ]
-then
-	# run the pstree report
-	run_pstree_report
+# run the pstree report
+if [[ ${USEPSTREE,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -734,53 +666,44 @@ create_output_dirs
...
 fi
 
-# check to see if netstat should be run
-if [ $USENETSTAT = "yes" -o $USENETSTAT = "YES" ]
-then
-	# run the netstat report
-	run_netstat_report
+# run the netstat report
+#TODO: standardize the run_netstat_report
+if [[ ${USENETSTAT,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -734,53 +666,44 @@ create_output_dirs
...
+  run_netstat_report
 fi
 
-# check to see if fdisk should be run
-if [ $USEFDISK = "yes" -o $USEFDISK = "YES" ]
-then
-	# run the fdisk report
-	run_fdisk_report
+# run the fdisk report
+if [[ ${USEFDISK,,} == "yes" ]]; then
```
StafDehat: Quotes

src/recap
```diff
@@ -734,53 +666,44 @@ create_output_dirs
...
 
-# check to see if mysqladmin should be run
+# run the mysql report
+#TODO: standardize the run_mysql_report
 type -p 'mysqladmin' > /dev/null
-if [[ $? -eq 0 ]] && [ $USEMYSQL = "yes" -o $USEMYSQL = "YES" ]
-then
-	# run the mysql report
-	run_mysql_report
+if [[ $? -eq 0 && ${USEMYSQL,,} == "yes" ]]; then
```
StafDehat: Quotes around USEMYSQL

src/recap
```diff
@@ -734,53 +666,44 @@ create_output_dirs
...
+  run_mysql_report
 fi
 
 # check to see if report should be emailed
-if [ ! $MAILTO = '' ]
-then
-	send_mail
+if [[ ! ${MAILTO} == '' ]]; then
```
StafDehat: Quotes 'round MAILTO

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
 }
 
-# print the output of "iotop" to the resources file
+# print the output of "iotop" to the specified file
 print_iotop() {
-	which iotop >/dev/null 2>&1 || return 1
-	echo "IOTOP report" >> $RESOURCES_FILE
-	iotop $OPTS_IOTOP >> $RESOURCES_FILE
+  local LOGFILE=$1
+  which iotop >/dev/null 2>&1 || return 1
```
tonyskapunk: to me seems like a remainder, I rather remove this or add to every single tool used,  ideally recap would be installed from a package(rpm or deb) in which case the dependency should be solved and  recap code should not double check that, for instance I've the list of package dependencies already sorted out in here https://github.com/tonyskapunk/recap/blob/recap_v1/util/packaging/rpm/recap.spec#L11 there is not a PR of this yet until this PR #78 and recaplog PR #72 are merged in master.  DEB packages are not in this repo, that's maintained outside the code, tagging @thebwt  as he is the packager for DEBs to see if the deb config is publicly available.  Finally on AUR I'm the maintainer of the package(currently recap-git) and is a dependency in there: https://aur.archlinux.org/packages/recap-git/

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
-        fi
+  local LOGFILE=$1
+  # We only support a single DOTMYDOTCNF for MySQL when running Plesk
+  local PLESK_FILE="/etc/psa/.psa.shadow"
+  if [[ -r "${PLESK_FILE}" ]]; then
+    echo "MySQL status" >> "${LOGFILE}"
+    mysqladmin -uadmin -p$(cat "${PLESK_FILE}") status >> "${LOGFILE}"
+    print_blankline "${LOGFILE}"
+  else
+    for cnf in $( sed -e 's/,/ /g' <<<"${DOTMYDOTCNF}" ); do
```
tonyskapunk: Not quite sure of this, to me works just fine, see that I'm allowing two/three ways to define the multiple list of configs:

1. White spaces
1. Comma separated values
1. Both

See this example:

Here the testing script, that mimics that for loop:
```bash
[root@gnu ~]# cat x
#!/bin/bash

multi_cnf() {
  multi_confs="$1"
  echo "processing: --${multi_confs}--"
  for cnf in $( sed -e 's/,/ /g' <<<"${multi_confs}" ); do
    if [[ -r "${cnf}" ]]; then
      echo "found: -${cnf}-"
    else
      echo "canot read -${cnf}-"
    fi
  done
  echo
}

# single
multi_cnf "/tmp/my.cnf0"
# multiple spaces
multi_cnf "/tmp/my.cnf0 /tmp/my.cnf1"
multi_cnf "/tmp/my.cnf0 /tmp/my.cnf1 /tmp/my.cnf2 /tmp/my_none"
multi_cnf " /tmp/my.cnf0  /tmp/my.cnf1   /tmp/my.cnf2   /tmp/my_none"
# multiple commas
multi_cnf "/tmp/my.cnf0,/tmp/my.cnf1"
multi_cnf "/tmp/my.cnf0,/tmp/my.cnf1,/tmp/my.cnf2,/tmp/my_none"
# multiple spaces and commas
multi_cnf " /tmp/my.cnf0, /tmp/my.cnf1,  /tmp/my.cnf2   /tmp/my.cnf3  ,/tmp/_my_none"
```

Creating some empty files
```bash
[root@gnu ~]# touch /tmp/my.cnf{0,1,2,3}
```

Now testing:
```bash
[root@gnu ~]# bash x
processing: --/tmp/my.cnf0--
found: -/tmp/my.cnf0-

processing: --/tmp/my.cnf0 /tmp/my.cnf1--
found: -/tmp/my.cnf0-
found: -/tmp/my.cnf1-

processing: --/tmp/my.cnf0 /tmp/my.cnf1 /tmp/my.cnf2 /tmp/my_none--
found: -/tmp/my.cnf0-
found: -/tmp/my.cnf1-
found: -/tmp/my.cnf2-
cannot read -/tmp/my_none-

processing: -- /tmp/my.cnf0  /tmp/my.cnf1   /tmp/my.cnf2   /tmp/my_none--
found: -/tmp/my.cnf0-
found: -/tmp/my.cnf1-
found: -/tmp/my.cnf2-
cannot read -/tmp/my_none-

processing: --/tmp/my.cnf0,/tmp/my.cnf1--
found: -/tmp/my.cnf0-
found: -/tmp/my.cnf1-

processing: --/tmp/my.cnf0,/tmp/my.cnf1,/tmp/my.cnf2,/tmp/my_none--
found: -/tmp/my.cnf0-
found: -/tmp/my.cnf1-
found: -/tmp/my.cnf2-
cannot read -/tmp/my_none-

processing: -- /tmp/my.cnf0, /tmp/my.cnf1,  /tmp/my.cnf2   /tmp/my.cnf3  ,/tmp/_my_none--
found: -/tmp/my.cnf0-
found: -/tmp/my.cnf1-
found: -/tmp/my.cnf2-
found: -/tmp/my.cnf3-
cannot read -/tmp/_my_none-
```

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
+    echo "MySQL processes" >> "${LOGFILE}"
+    if [[ ${MYSQL_PROCESS_LIST} == "table" ]]; then
+      mysqladmin -uadmin -p$( cat "${PLEASK_FILE}" ) -v processlist >> "${LOGFILE}"
+    fi
+    if [[ ${MYSQL_PROCESS_LIST} == "vertical" ]]; then
+      mysql -uadmin -p$( cat "${PLESK_FILE}" ) -e "show full processlist\G" >> "${LOGFILE}"
+    fi
+    print_blankline "${LOGFILE}"
+  else
+    for cnf in $( sed -e 's/,/ /g' <<<"${DOTMYDOTCNF}" ); do
```
tonyskapunk: should be similar to https://github.com/rackerlabs/recap/pull/78#discussion_r105439236

src/recap
```diff
@@ -146,492 +134,440 @@ print_usage() {
...
-	done
+  local LOGFILE=$1
+  # First, we establish which tmpdir and pid_file are in use, strip any
+  # trailing slash and populate the file with current information.
+  # We throw away the output of the "show engine innodb status" command as
+  # it is likely being truncated, which we attempt to work around:
+
+  echo "MySQL InnoDB status"  >> "${LOGFILE}"
+  unset MYVALS PID_FILE TMPDIR
+  MYVALS=( $( IFS=$'\t' mysql -Bse "SHOW VARIABLES LIKE 'pid_file'; \
```
tonyskapunk: There are three functions for mysql interactions, only two of them loop through multiple mysql configuration: `print_mysql`, `print_mysql_procs`, but not for `print_mysql_innodb_status` as you pointed out.  All these functions are being called from: `run_mysql_report` this makes me think if we should instead of looping inside each `print` function or from the `run` function, need to analyze this better and take a decision, this will come in a separate commit.

StafDehat: That works.  Logged the to-do item here https://github.com/rackerlabs/recap/issues/81

src/recap
StafDehat: You're right, in the context of how you read my comment.  Apologies, I didn't communicate the issue very well.  The potential "whitespace problem" is if you've got spaces in the filepath of the my.cnf file.  If the filepath has whitespace, then you'll handle one filepath as two iterations.  The typical fix for *that* whitespace issue would be to quote it, but since for-loops use whitespace as a separator, you'd break your loop by quoting the "variable".

This code illustrates the problem I'm trying to convey:
```
ahoward@galapagos ~ $ multi_confs="/etc/mysql config.cnf"
ahoward@galapagos ~ $ multi_confs="${multi_confs},/etc/my.cnf"
ahoward@galapagos ~ $ for cnf in $( sed -e 's/,/ /g' <<<"${multi_confs}" ); do
> echo "${cnf}"
> done
/etc/mysql
config.cnf
/etc/my.cnf
ahoward@galapagos ~ $ for cnf in "$( sed -e 's/,/ /g' <<<"${multi_confs}" )"; do echo "${cnf}"; done
/etc/mysql config.cnf /etc/my.cnf
ahoward@galapagos ~ $ 
```
Notice, the first for-loop iterates 3x times, and iterations 1 & 2 are each broken due to the partial filename.  The typical fix for that whitespace issue is the quote the var, but that's not an option in for-loops, as illustrated by the second example above.

That's why I'd recommended using "while read LINE; do ... done < <(tr ',' '\n')", or arrays, or just mandate a single mysql-defaults-file instead of trying to handle multiple.

tonyskapunk:
Most of the feedback has been addressed, only pending piece is around mysql functions, already documented in here #81

src/recap
tonyskapunk: This is why I'm using it:

example code:
```bash
[root@gnu ~ ]# cat y
#!/bin/bash

items="foo bar baz"
use_foo=FOO
use_bar=BAR
use_baz=BAZ

echo not using eval
for item in ${items}; do
  item_use=$( echo "\${use_${item}}" )
  echo -${item_use}-
done

echo using eval
for item in ${items}; do
  item_use=$( eval "echo \${use_${item}}" )
  echo -${item_use}-
done
```

Output of the tests:
```bash
[root@gnu ~]# bash y
not using eval
-${use_foo}-
-${use_bar}-
-${use_baz}-
using eval
-FOO-
-BAR-
-BAZ-
```

src/recap
tonyskapunk: Ah, I see your point now, that seems to be valid, somehow unusual but valid as any file can have white spaces, I'd definitely cover this use case as long as mysql does it as well via its config file, and it does:
```bash
[root@gnu ~]# mysqladmin --defaults-extra-file="/root/my config.cnf" ping
mysqld is alive
```

In that case we will need to move to support only comma separated values and not white spaces, considering the white spaces are part of a name, e.g.

* Valid config:
  - DOTMYDOTCNF="/root/.my.cnf,/root/.my2.cnf,/some/path with/spaces in name.cnf"
  - DOTMYDOTCNF="  /some/path/my.cnf , /some/other/my.cnf "  (in this case we can remove trailing spaces)
* Invalid config:
  - DOTMYDOTCNF="/some/path/my.cnf /some/other/my.cnf"

Or not support it at all.

src/recap
StafDehat: Just do this instead:
```
ITEM_USE="USE${item^^}"
```

tonyskapunk: Nice, ended up being:

```bash
echo not using eval
for item in ${items}; do
  item_use="use_${item}"
  echo -${!item_use}-
done
```

For a indirect reference, obtaining the desired:
```bash
# bash y                                                                                                                        
not using eval
-FOO-
-BAR-
-BAZ-
using eval
-FOO-
-BAR-
-BAZ-
```

tonyskapunk:
Last commit is to avoid using `eval` on variable name of variables.

tonyskapunk:
Last commit is a fix for #80 
- Removing support for multiple my.cnf files
- Better lock implementation

src/recap
```diff
@@ -283,49 +282,47 @@ print_mysql_innodb_status() {
...
+      if [[ -r "${cnf}" ]]; then
         echo "MySQL ${cnf} processes" >> "${LOGFILE}"
         if [[ ${MYSQL_PROCESS_LIST} == "table" ]]; then
           mysqladmin --defaults-extra-file="${cnf}" -v processlist >> "${LOGFILE}"
-        fi
-        if [[ ${MYSQL_PROCESS_LIST} == "vertical" ]]; then
+        elif [[ ${MYSQL_PROCESS_LIST} == "vertical" ]]; then
           # This doesnt seem to work right
-          mysqladmin --defaults-extra-file="${cnf}" --vertical processlist
+          mysqladmin --defaults-extra-file="${cnf}" -e "show full processlist\G"
```
StafDehat: Still broken: s/mysqladmin/mysql/
That being said, you fixed it for real in https://github.com/rackerlabs/recap/pull/78/commits/d1d27cdbce58069d099e902fca8c33189ab594a8, so nevermind.

src/recap
```diff
@@ -576,25 +569,24 @@ check_disk_space() {
...
 	echo "No configuration file found. Proceeding with defaults."
 else
 	source /etc/recap
 fi
 
 # Quit if load is higher than MAXLOAD
-LOAD=$( uptime | awk -F'load average:' '{ print $2 }' |
-          cut -d" " -f2 | cut -d"," -f1 )
-QUIT=$( echo ${LOAD}'>'${MAXLOAD} | bc -l )
+LOAD=$( grep -Po '^[\d.]+' /proc/loadavg )
```
StafDehat: This does actually seem to work, but I'm surprised at that.  "." is supposed to match 1 of any character, so I'd expect this regex to match the whole file, not just the first loadavg entry.  Demo:
```
galapagos ~ # grep -Po '.+' /proc/loadavg
1.46 1.40 1.13 1/854 2847
galapagos ~ #
```
The regex as it's currently written does actually work how we want, but it seems to violate what I know about regex.  I'd feel more comfortable with this if we escaped the '.' - please do?

src/recap
```diff
@@ -640,90 +569,85 @@ check_disk_space() {
...
 fi
 
 # Quit if load is higher than MAXLOAD
-LOAD=$(uptime | awk -F'load average:' '{ print $2 }' | cut -d" " -f2 | cut -d"," -f1)
-QUIT=$(echo $LOAD'>'$MAXLOAD | bc -l)
-if [ $QUIT -eq 1 ]
-then
-    echo "QUIT (LOAD:"$LOAD", MAXLOAD:"$MAXLOAD")"
-    exit 1
+LOAD=$( grep -Po '^[\d.]+' /proc/loadavg )
```
StafDehat: This does actually seem to work, but I'm surprised at that.  "." is supposed to match 1 of any character, so I'd expect this regex to match the whole file, not just the first loadavg entry.  Demo:
```
galapagos ~ # grep -Po '.+' /proc/loadavg
1.46 1.40 1.13 1/854 2847
galapagos ~ #
```
The regex as it's currently written does actually work how we want, but it seems to violate what I know about regex.  I'd feel more comfortable with this if we escaped the '.' - please do?

StafDehat: That's it, though.  Good work.  :+1: , but please do escape that '.'

tonyskapunk: Inside square brackets(character sets) the `.` represents a `"dot"` the purpose of the square bracket is to match a single char with a specific pattern, examples
-  `[\d]` a digit
- `[a-z]` a range
- `[YyNn]` a y/n option
- `[etc]` - `e` or `t` or `c` :P

I agree at first sight seems confusing, but is valid, I'll make the change.

See this description of the perl regex flavor syntax for the character sets, while is not explicit, it is implied that the use of `[]` is to evaluate matching patters to a single char, if `[.]` meant any char, the purpose of its use would be null as that can be written `.`. http://www.boost.org/doc/libs/1_50_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html#boost_regex.syntax.perl_syntax.character_sets

There shouldn't be any issue making the change:

Matching a dot or more:
```bash
  $ echo -e "1.23\n1_23\n1/23\n1.2.3" | grep -P "[.]+"
  1.23
  1.2.3

  $ echo -e "1.23\n1_23\n1/23\n1.2.3" | grep -P "[\.]+"
  1.23
  1.2.3
```

tonyskapunk:
Latest commit is explicit in the regex :)

StafDehat:
:+1:

tonyskapunk:
Added 9dd06a8 to address #86

tonyskapunk:
@StafDehat completely agree with your point, I missed the fact that args could be changed, I removed the mentioned commit.

tonyskapunk:
Added new commit 02e12c6 , this one simply includes stderr into stdout to prevent errors, but still logs the Warning message about GPT into the logs. Fix #86

StafDehat:
I like it. :+1:

kevi5702:
:+1: :dancer:

PR #79, opened 2017-02-10
Adding recaptool and better path handling.

tonyskapunk:


Makefile
```diff
@@ -1,43 +1,52 @@
...
 	@echo "Installing man pages..."
-	@install -Dm0644 src/recap.5 $(MANDIR)/man5/recap.5
-	@install -Dm0644 src/recap.8 $(MANDIR)/man8/recap.8
-	@install -Dm0644 src/recaplog.8 $(MANDIR)/man8/recaplog.8
+	@install -Dm0644 src/recap.5 $(DESTDIR)$(MANDIR)/man5/recap.5
+	@install -Dm0644 src/recap.8 $(DESTDIR)$(MANDIR)/man8/recap.8
+	@install -Dm0644 src/recaplog.8 $(DESTDIR)$(MANDIR)/man8/recaplog.8
+	@install -Dm0644 src/recaptool.8 $(DESTDIR)$(MANDIR)/man8/recaptool.8
 	@echo "Installing configuration..."
-	@install -Dm0644 src/recap.conf $(SYSCONFDIR)/recap
```
troyengel: My only question is this one -- why are we stripping the dot-conf off the end of the file when installing it? Is this part of expected operation?

troyengel: oops tagged one line in the PR too high, I meant the next line down sorry.

thebwt: according to https://github.com/rackerlabs/recap/blob/master/src/recap#L68 That's the proper name of the config file

troyengel: *sigh* OK - changing that would break a simple upgrade, we should really be naming that file recap.conf. :(

tonyskapunk: So yes, as @thebwt points out is what's used on recap.  I chase whether or not this is a good practice or not and could've swear I found a source that would make a distinction on that matter, something like this (from what I remember):

- /etc/__project__  - OK when there is no other config file
- /etc/__project.conf__ - OK when there is a directory file for that project and requires more configs, e.g. /etc/__project__/
- /etc/__project__/ - From above, when needed to include more configs
- /etc/__project.d__/ - Similar as above but more explicit, when needed to include more configs

Don't remember if it was a best practice for packaging or something similar but that made me think it was not completely wrong.  But truth is I don't find at the moment anything to support that and since I don't see any reference to this in the [FSH for /etc/](http://www.pathname.com/fhs/pub/fhs-2.3.html#ETCHOSTSPECIFICSYSTEMCONFIGURATION)  this opens the question to.  What's the right thing to do?

For now is clear that without a new release a change like this may not be recommended, but definitely something to take into consideration.

thebwt: I agree, this PR is not the place for a change like that.

tonyskapunk: alright, I've created #80 to keep the discussion of the config file, since there is no other problem with the Makefile I'll proceed to merge it.

troyengel: 👍

tonyskapunk:
Merging this one.

PR #82, opened 2017-03-24
Update Makefile - make DOCDIR so install succeeds

jamrok:


Makefile
```diff
@@ -26,6 +26,7 @@ install:
 	@echo "Installing cron job..."
 	@install -Dm0644 src/recap.cron $(DESTDIR)$(CRONDIR)/recap
 	@echo "Installing docs..."
+	@install -dm0755 $(DESTDIR)$(DOCDIR)/recap
```
carlwgeorge: This change isn't necessary, because the very next line calls `install` with a `-D` flag, which creates the target directory for you.

jamrok: I thought so too, but it's failing to do so:
```
# echo $(cat /etc/issue; uname -r)
Ubuntu 14.04.4 LTS \n \l 3.13.0-95-generic

# make install
Installing scripts...
Installing man pages...
Installing configuration...
Installing cron job...
Installing docs...
install: failed to access ‘/usr/local/share/doc/recap/’: No such file or directory
make: *** [install] Error 1
```
It was only fixed with that addition

carlwgeorge: How strange, I'll take another look.

tonyskapunk: @jamrok Thanks for reporting this, I was able to reproduce the error:

```bash
root@docker:/recap# cat /etc/os-release; uname -r
NAME="Ubuntu"
VERSION="14.04.5 LTS, Trusty Tahr"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 14.04.5 LTS"
VERSION_ID="14.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
3.10.0-514.10.2.el7.x86_64

root@docker:/recap# make install
Installing scripts...
Installing man pages...
Installing configuration...
Installing cron job...
Installing docs...
install: failed to access '/usr/local/share/doc/recap': No such file or directory
make: *** [install] Error 1
```

Digging around I found the problem is the underlying directory that does not exist:

```bash
root@docker:/recap# ls -ld /usr/local/share/doc
ls: cannot access /usr/local/share/doc: No such file or directory

root@docker:/recap# touch TEST
root@docker:/recap# install -D TEST -t /usr/local/share/doc/recap 
install: failed to access '/usr/local/share/doc/recap': No such file or directory
```

Comparing the manual of `install` between 14.04 and 16.06 I see differences:

```bash
## 14.04
root@docker:/recap# man install | grep -- -D
       -D     create all leading components of DEST except the last, then copy SOURCE to DEST

## 16.06
root@docker:~# man install | grep --  -D
       -D     create all leading components of DEST except the last, or all components of --target-directory, then copy SOURCE to DEST
```

Your patch should solve this problem on `14.04` but interestingly enough there is no package in the path in question: `/usr/local/share/doc/`

```bash
root@docker:/recap# apt-file search /usr/local/share/doc/
root@docker:/recap#
```

Nor mention of it in the  [LFH](http://www.tldp.org/LDP/Linux-Filesystem-Hierarchy/html/usr.html).

---

Even when this might not be a problem in newer Ubuntu versions or other distros it does not hurt having the extra line to create it. I'm good with the proposed change.

:+1:

tonyskapunk:
As an additional note, `PREFIX` and `DEST` are usually  modified by the packages to obtain the desired result, here is an example on how `PREFIX` is modified in arch linux: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=recap-git#n26

Since is a variable it can simply be overwritten in your make command like this:
```bash
make PREFIX=/usr install
```

tonyskapunk:
@jamrok please remove the last commit so we can merge this, the `cron` problem is fixed in #84 as reported in #83

tonyskapunk:
@justfortier submitted #85 to fix the install issue in U14.04, @jamrok please drop 726b77c from this commit as discussed so it can be fixed for both.

Issue #90, opened 2017-05-16
recap installer vs makefile


tonyskapunk:
Seems a duplicated effort having two scripts that produce almost the same outcome, here the differences between them:

1. `Recap-installer` is more explicit with directory permissions than `Makefile`:
  ```log
  | Directory                | recap-installer | Makefile |
| /var/log/recap           |            0700 |     0750 |
| /var/log/recap/backups   |            0700 |     0750 |
| /var/log/recap/snapshots |            0700 |     0750 |
```
2. The `recap-installer` places the docs in a [directory dependent of the version](https://github.com/rackerlabs/recap/blob/master/util/recap-installer#L54-L56), The [makefile is generic](https://github.com/rackerlabs/recap/blob/master/Makefile#L30-L31).
1. The `recap-installer` does not include `recaptool.8` man page
1. The `recap-installer` [gzips the man pages](https://github.com/rackerlabs/recap/blob/master/util/recap-installer#L61-L63), while the `Makefile` does not, but for instance the [RPM spec takes care of this](https://github.com/rackerlabs/recap/blob/master/util/packaging/rpm/recap.spec#L47-L49).
1. The `recap-installer` provides support to [macosx for the cronjob](https://github.com/rackerlabs/recap/blob/master/util/recap-installer#L61-L63), `Makefile` does not.
1. The `recap-installer` looks for the [httpd config directory](https://github.com/rackerlabs/recap/blob/master/util/recap-installer#L76-L81) if found the [recap.httpd.conf](https://github.com/rackerlabs/recap/blob/master/src/recap.httpd.conf) file is copied over, this is disabled by default, i.e commented out.
1. `Makefile` is more flexible than `recap-installer` on the path to be installed.

carlwgeorge:
IIRC the original intent was for the Makefile to obsolete the installer script.  Some extra notes:

* Regarding 2, check out https://fedoraproject.org/wiki/Changes/UnversionedDocdirs.  That only applies to Fedora, and RHEL/EPEL packages still use the old versioned style.  It is handled automatically by the `%doc` and `%license` macros, so we don't have to adjust for it in the spec file.  But that makes me realize that from a packaging perspective we don't want the docs installed by `make install`, as we handle them with those macros.  I'll open a separate issue to discuss creating a standalone `doc` target in the Makefile.
* Regarding 4, check out https://fedoraproject.org/wiki/Packaging:Guidelines#Manpages for the Fedora guideline on this.

[...]

PR #89, opened 2017-05-15
Release 1.0.0

tonyskapunk:

PR #91, opened 2017-05-17
multiple makefile targets

carlwgeorge:
In RPM spec files, putting doc files in place is handled by the %doc and %license macros.  In other words, we need to be able to explictly install everything _but_ the docs.  This change preserves the existing behavior of `make install`/`make uninstall`, but splits the actions out into separate targets.  This will allow us to only run `make install-base` and `make install-man` in the spec file.  Additionally, this change templatizes the recap.cron file and generates it appropriately.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,46 +9,71 @@ CRONDIR       ?= $(SYSCONFDIR)/cron.d
 LOGDIR        ?= /var/log
 
 all:
-	@echo "Nothing to compile, run make install."
+	@echo "Nothing to compile, please choose a specific target:"
+	@echo "    install (includes install-base, install-man, and install-doc)"
+	@echo "    install-base"
+	@echo "    install-man"
+	@echo "    install-doc"
+	@echo "    uninstall (includes uninstall-base, uninstall-man, and uninstall-doc)"
+	@echo "    uninstall-base"
+	@echo "    uninstall-man"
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -9,46 +9,68 @@ CRONDIR       ?= $(SYSCONFDIR)/cron.d
...
+	@echo "    install-man"
+	@echo "    install-doc"
+	@echo "    uninstall (includes uninstall-base, uninstall-man, and uninstall-doc)"
+	@echo "    uninstall-base"
+	@echo "    uninstall-man"
+	@echo "    uninstall-doc"
 
-install:
+src/recap.cron:
+	@sed -e 's|@BINDIR@|$(BINDIR)|' src/recap.cron.in > src/recap.cron
```
tonyskapunk: As a note, if this is merged it will break the `recap-installer` in #90 the attempt is to remove the installer as the `Makefile` should be able to replace it. Mentioning this as we need coordination to prevent breaking the installer.

tonyskapunk:
It may be worth looking into targeting these changes to the PR #89 as they are related, mainly on the `spec` file.

carlwgeorge:
In order to cleanly integrate this with [milestone 1.0](https://github.com/rackerlabs/recap/milestone/1), I'm going to submit it as a PR to tonyskapunk:recap_v1 instead.

PR #92, opened 2017-05-22
Fixed apache fullstatus

mik11231:
There was a missing variable for which log to send fullstatus to.
Added variable ${ITEM_FILE} following the print_httpd_fullstatus command to resolve the issue.


tonyskapunk:
ah great catch @mik11231, thanks for fixing it!

PR #95, opened 2017-06-09
add GPL2 notice to source files

carlwgeorge:
I believe this is necessary to properly comply with the license.


tonyskapunk:
lgtm

PR #99, opened 2017-06-13
Removing specific permissions for the recap logs.

### LINKED ISSUE #98: 750 log directory permissions
I'm working on packaging recap for Fedora/EPEL.  Rpmlint complains about the 750 permissions on the log directories.

```
recap.noarch: E: non-standard-dir-perm /var/log/recap 750
recap.noarch: E: non-standard-dir-perm /var/log/recap/backups 750
recap.noarch: E: non-standard-dir-perm /var/log/recap/snapshots 750
```

These were set this way in [2012](https://github.com/rackerlabs/recap/blob/1.0.0/CHANGELOG.md#092---2012-11-01).  Are these permissions actually necessary?  Is any recap output actually sensitive?

References:
* https://github.com/rackerlabs/recap/blob/1.0.0/Makefile#L42-L44
* https://github.com/rackerlabs/recap/blob/1.0.0/src/recap#L88-L89
* https://github.com/rackerlabs/recap/blob/1.0.0/CHANGELOG.md#092---2012-11-01

tonyskapunk:
Fix #98 

Setting regular permissions 755 instead, as there is no information in the `recap` logs that requires to be kept secret.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,8 +57,8 @@ install-base: src/recap.cron
 	@install -Dm0644 src/recap.cron $(DESTDIR)$(CRONDIR)/recap
 	@echo "Creating log directories..."
 	@install -dm0750 $(DESTDIR)$(LOGDIR)/recap
-	@install -dm0750 $(DESTDIR)$(LOGDIR)/recap/backups
-	@install -dm0750 $(DESTDIR)$(LOGDIR)/recap/snapshots
+	@install -dm0755 $(DESTDIR)$(LOGDIR)/recap/backups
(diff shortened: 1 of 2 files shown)


piersc:
recap can record things that wouldn't otherwise be visible to a non-superuser, e.g. the process associated with a connection in netstat output or the MySQL process list. For this reason the logs probably shouldn't be visible to non-root.

tonyskapunk:
Finally this keeps the 0750 perms on `BASEDIR` default is `/var/log/recap` directories inside are now 0755 instead of 0750

PR #116, opened 2017-09-08
Information on the locations using make

tonyskapunk:
Updating information about the default locations through make, including examples.


tonyskapunk:
Merging, as this is an update on the documentation.

PR #119, opened 2017-09-15
Standardizing the help on all the scripts.

### LINKED ISSUE #112: Standardize help/usage
Ensure all the scripts use the same way to produce the help/usage via CLI, I'm bias in this aspect as I prefer the use of embedded help in the comments as in `recaplog`: [here](https://github.com/rackerlabs/recap/blob/master/src/recaplog#L39-L45) and [here](https://github.com/rackerlabs/recap/blob/master/src/recaplog#L257-L259)

tonyskapunk:
- Externally there are just minor changes to the output
- Internally all of them behave the same way
- Included some missing exit codes

Edit:
Fix #112

### DIFF
--- a/src/recap
+++ b/src/recap
@@ -39,6 +39,14 @@
 # Define the default environment for recap
 # DO NOT change these values here, user-modifiable options can be tweaked through
 # /etc/recap
+#
+#~ Usage: _tool_ [OPTION]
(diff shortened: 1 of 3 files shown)


tonyskapunk:
Merging into development after a week.

PR #122, opened 2017-09-29
Deprecating netstat in favor of ~modern~ current tools

### LINKED ISSUE #107: Make use of newer tools for networking stats
Most of the network stats depend on old `net-tools`, while this provides compatibility it would be desired to use more modern alternatives.

tonyskapunk:
Fix #107 

- Deprecates the use of `netstat` to generate reports
- The replacements are: `nstat` and `ss` both provided through `iproute2` utilities (Red Hat names the package of this utilities `iproute`)
- Log report file does **not** change, remains being: `${BASEDIR}/netstat_<timestamp>.log`
- Options and Command options **names** remain the **same**:
  - `USENETSTAT`, `USENETSTATSUM`.
  - `OPTS_NETSTAT`, `OPTS_NETSTAT_SUM`.
- Default values have changed **only** for command options:
  - `OPTS_NETSTAT` - `-atunp`
  - `OPTS_NETSTAT_SUM` - `-a`
- Content of the reports change due to the nature of the new tools, examples of the old vs new are included in this gist: https://gist.github.com/tonyskapunk/f343648806d5d1db279aad1a7958b666

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,14 +162,14 @@ These are the type of reports generated and their dependencies.
 
 ##### netstat
 
-- **USENETSTAT** - Generates network stats from "netstat `${OPTS_NETSTAT}`"
+- **USENETSTAT** - Generates network stats from "ss `${OPTS_NETSTAT}`"
 
   Required by: `USENETSTATSUM`
 
   Default: `USENETSTAT="yes"`
 
 
-- **USENETSTATSUM** - Generates logs from "netstat `${OPTS_NETSTAT_SUM}`".
+- **USENETSTATSUM** - Generates logs from "nstat `${OPTS_NETSTAT_SUM}`".
 
   Requires: `USENETSTAT`
 
@@ -291,17 +291,17 @@ Options used by the tools generating the reports
 
   Default: `OPTS_IOTOP="-b -o -t -n 3"`
 
-- **OPTS_NETSTAT** - netstat options
+- **OPTS_NETSTAT** - ss options
 
   Required by: `USENETSTAT`
 
-  Default: `OPTS_NETSTAT="-ntulpae"`
+  Default: `OPTS_NETSTAT="-atunp"`
 
-- **OPTS_NETSTAT_SUM** - netstat statistics options
+- **OPTS_NETSTAT_SUM** - nstat options
 
   Required by: `USENETSTATSUM`
 
-  Default: `OPTS_NETSTAT_SUM="-s"`
+  Default: `OPTS_NETSTAT_SUM="-a"`
 
 - **OPTS_PS** - ps options
 
--- a/src/recap
+++ b/src/recap
@@ -58,54 +58,89 @@ declare -r LOG_SUFFIX=$( date +%Y%m%d-%H%M%S )
 PATH=/bin:/usr/bin:/sbin:/usr/sbin
 
 ## Default settings(modified through command line arguments)
-BACKUP="no"
-SNAPSHOT="no"
+declare -r default_BACKUP="no"
+declare -r default_SNAPSHOT="no"
+BACKUP="${default_BACKUP}"
+SNAPSHOT="${default_SNAPSHOT}"
 
 ## Default settings(can be overridden in config file)
-BACKUP_ITEMS="fdisk mysql netstat ps pstree resources"
-BASEDIR="/var/log/recap"
-MAILTO=""
-MIN_FREE_SPACE=0
-USEFDISK="no"
-USEPS="yes"
-USEPSTREE="no"
-USESLAB="no"
+declare -r default_BACKUP_ITEMS="fdisk mysql netstat ps pstree resources"
+declare -r default_BASEDIR="/var/log/recap"
+declare -r default_MAILTO=""
+declare -r default_MIN_FREE_SPACE=0
+declare -r default_USEFDISK="no"
+declare -r default_USEPS="yes"
+declare -r default_USEPSTREE="no"
+declare -r default_USESLAB="no"
+BACKUP_ITEMS="${default_BACKUP_ITEMS}"
+BASEDIR="${default_BASEDIR}"
+MAILTO="${default_MAILTO}"
+MIN_FREE_SPACE=${default_MIN_FREE_SPACE}
+USEFDISK="${default_USEFDISK}"
+USEPS="${default_USEPS}"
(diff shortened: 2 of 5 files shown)


troyengel:
Posing the question - since we're doing all this work to refactor for the new tools, why not rename the Options and Commands (ala ```USENSTAT```, ```USENSTATSUM```, ```OPS_NSTAT``` etc.) while we're here? And rename the logfile from netstat to nstat? My thoughts:

* it breaks backwards compatibility anyways, as the contents of the file is not the same as the old contents; the log files are not the same, so we shouldn't try and treat them the same -- the format of nstat/ss is vastly different than netstat and would break any script trying to use the logs
* if the user has configured options that work with netstat and not nstat (same with ss), then this ensures they don't end up with output that's broken? For example, netstat has ```--numeric-ports``` that when uses with nstat causes a crash; so an upgrade without the user's config being upgraded would result in the user getting *no* metrics for nstat

I kind of think that the update should be branded as a backwards-breaking change (let's call it 2.0) and have needs the user upgrade some config options as needed. $0.02!

carlwgeorge:
Changing a log file format would be acceptable for an EPEL update.  Renaming a log file would probably also be fine.  However, an [update that requires manual intervention would not be OK](https://fedoraproject.org/wiki/EPEL_Updates_Policy).  If we go the route of "recap 2.0" then I would only be able to do that update for Fedora branches, not EPEL.

If a user has configured `netstat` options that don't work with `ss`, then ideally recap should ignore the user configured options and fallback to the default.

tonyskapunk:
@troyengel I thought a while about the log name(`netstat_*.log`), its options(`USENETSTAT`, `USENETSTATSUM`, etc) and their  command options(`OPTS_NETSTAT`, `OPTS_NETSTAT_SUM`, etc.) at the end got to the conclusion that it would be OK to remain unchanged.  They are "Network Statistics" hence `netstat_*.log`.

Internally  the logs produced by a command(or a set of commands) can be tweaked by their options and I've ensured the documentation is accurate and up-to-date to prevent misunderstandings.

The change is definitely not backward-compatible, but it allows some upgrade without breaking horrendously.
  - **nstat** - This one is used to produce the "Network Stats Summary Report" its default is `USENETSTATSUM="no"`, the default option was `OPTS_NETSTAT_SUM="-s"` now is `OPTS_NETSTAT_SUM="-a"`.  If someone has this summary enabled and somehow the old default config remains, it will not fail, it will simply produce a different output. (`-s, --noupdate`, this produces accumulated values from the last measurement). Of course if other arguments are added in the config, those could prevent nstat to run. (The only possible options should be: `netstat {--statistics|-s} [--tcp|-t] [--udp|-u] [--udplite|-U] [--sctp|-S] [--raw|-w] [delay]`). **Note** that `-n` is not an option for the summary.

  - **ss** - This one is used to produce the "Network Stats Report" its default is `USENETSTAT="yes"`, de default option was `OPTS_NETSTAT="-ntulpae"` now is `OPTS_NETSTAT="-atunp"` .  If somehow the old default settings remain, the command will not fail, just different output.  

---

In summary:
  - Old default configurations will not break the changes introduced by this PR.
  - Old custom configurations could potentially break the output with the new commands.

From the  [versioning](https://github.com/rackerlabs/recap/blob/development/README.md#versioning) perspective this could be seen as a  change that don't break any rolling upgrades but require closer user attention to configuration defaults, function behavior, and output formatting.  Making it candidate for **1.2.0**

---

@carlwgeorge from the package perspective, it sounds like as long as we ensure that any wrong config detected is logged and goes to the default one instead, is that right?

carlwgeorge:
Sounds good to me.

tonyskapunk:
Last commit takes care of the validation of the settings in the `netstat_*.log` reports(for both `ss` and `nstat`) in case there is an invalid setting(based on exit code) it will use the default settings.

Some examples:

## Bad config:

```bash
$ sed -e 's/^\s*#.*//' -e '/^$/d' /etc/recap 
USENETSTATSUM="yes"
OPTS_NETSTAT="-ntulpae --verybad"
OPTS_NETSTAT_SUM="-s --bad"
```

`recap` completes successfully using the default settings instead, the logs contains hints of the issue:

```bash
$ sudo head -n1 /var/log/recap/netstat_20171027-115936.log; sudo grep -n -C1 -P "^Bad|^Network" /var/log/recap/netstat_20171027-115936.log
2017-10-27_11:59:36
1-2017-10-27_11:59:36
2:Network connections
3-Netid  State      Recv-Q Send-Q Local Address:Port               Peer Address:Port              
--
20-
21:Bad config: '-ntulpae --verybad',  using default: '-atunp'
22- 
23:Network traffic summary
24-#kernel
--
124-
125:Bad config: '-s --bad',  using default: '-a'
```

## Old Config

```bash
$ sed -e 's/^\s*#.*//' -e '/^$/d' /etc/recap 
USENETSTATSUM="yes"
OPTS_NETSTAT="-ntulpae"
OPTS_NETSTAT_SUM="-a"
```
`recap` completes successfully without the need of using default settings as the **old** settings don't break the new commands.

```bash
$ sudo head -n1 /var/log/recap/netstat_20171027-120634.log; sudo grep -n -C1 -P "^Bad|^Network" /var/log/recap/netstat_20171027-120634.log
2017-10-27_12:06:34
1-2017-10-27_12:06:34
2:Network connections
3-Netid  State      Recv-Q Send-Q Local Address:Port               Peer Address:Port              
--
22- 
23:Network traffic summary
24-#kernel
```

---

@troyengel @carlwgeorge  feel free to review and provide feedback/comments, Thanks!

troyengel:
Neat idea, and I see you prepared the rest of the code to implement this methodology on a lot of the other commands as well later. I took a look at the ss/nstat [source code](https://git.kernel.org/pub/scm/linux/kernel/git/shemminger/iproute2.git/tree/misc) and see all their exits are generally the same (0, 1 or -1), good to go for me to solve the backwards compatibility mishaps.

tonyskapunk:
@troyengel correct, at this point is only evaluating the commands that produces the `netstat.*.log` reports but eventually it will be tested each one of those.

tonyskapunk:
Merging this one into `development` I want to get this rolling and prepare a new PR for releasing a new version, I'll tag anyone related for reviewing before release.

PR #130, opened 2017-10-27
Deprecating MAXLOAD, adding timeout.

### LINKED ISSUE #123: Increase default MAXLOAD
Hey there,

I see in #100 we made changes to the default MAXLOAD variable:
> NEW: MAXLOAD - 10*cpu(dynamic by default, was 1000)
> OLD: MAXLOAD="1000"

This had previously been set over in 4c03a4c8b7f9bd6674a47f85d90c07e973e83edc with the logic that we'd rather have logging in the event of an issue and further crash than massive empty gaps where we are left guessing what might have been going on.

While it is agreeable that `recap` and the tools it triggers will contribute to the load, the "black box" style information of why we crashed is preferred default. I've spoken with Man Chung and he still believes that  this value is too low - though he might chime in with more words.

tonyskapunk:
- Deprecated the use of `MAXLOAD`.
- Timeout added to cronjobs with a default of 5m for most cases.

---

If any old configuration is making use of `MAXLOAD`, this will be simply ignored.

Fix #123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
-## [unreleased]
+## [Unreleased]
 - Deprecated the use of OPTS_CURL in favor of OPTS_LINKS
 - Default changed for OPTS_STATUSURL
+- Deprecated the use of MAXLOAD. [(#123)][1]
+- Timeout added to cronjobs with a default of 5m for most cases. [(#123)][1]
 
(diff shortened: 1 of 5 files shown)


man-chung:
Looks ok

tonyskapunk:
Thanks for looking @man-chung I'm going to rebase and give some time for any other reviewer

LukeHandle:
LGTM :+1:

tonyskapunk:
merging into `development`

PR #132, opened 2017-11-02
recap release 1.2.0

tonyskapunk:
- Use of modern tools for netstat reports, `net-tools` are now deprecated. (#107)
  - https://bugzilla.redhat.com/show_bug.cgi?id=1496151
- Deprecated the use of `OPTS_CURL` in favor of `OPTS_LINKS` (#125)
- Default changed for `OPTS_STATUSURL` (#125)
- Deprecated the use of `MAXLOAD`. (#123)
- Timeout added to cronjobs with a default of 5m for most cases. (#123)
- Dependencies:
  - Removed:`bc`, `net-tools`.
  - Added: `links`, `iproute`/`iproute2`.

---

cc for review and/or comments: @carlwgeorge @man-chung  @terackspace @LukeHandle @Vascom

PR #133, opened 2017-11-02
Adding quotes to support white-spaced paths.

tonyskapunk:
Per comments in #132 fixing the missing quotes on a variable used for a path.


terackspace:
👍

PR #137, opened 2017-11-10
Pre-decommission of /etc/recap in favor of /etc/recap.conf.

### LINKED ISSUE #80: Config file location
Based on the discussion on #79 revisit the location for the configuration file, currently in `/etc/recap` the discussion in that PR suggests that it might be better to have it under `/etc/recap.conf` we can discuss the pros, cons in this issue rather that on the PR.

tonyskapunk:
These changes should slowly(through one release) warn about the lack of the new location of the config:
-  `/etc/recap.conf`

In case that's missing the old location is read `/etc/recap`, the old location has precedence over the new location for at least one release.

Fix #80

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ install-base:
 	@install -Dm0755 src/recaplog $(DESTDIR)$(BINDIR)/recaplog
 	@install -Dm0755 src/recaptool $(DESTDIR)$(BINDIR)/recaptool
 	@echo "Installing configuration..."
-	@install -Dm0644 src/recap.conf $(DESTDIR)$(SYSCONFDIR)/recap
+	@install -Dm0644 src/recap.conf $(DESTDIR)$(SYSCONFDIR)/recap.conf
 	@echo "Creating log directories..."
 	@install -dm0750 $(DESTDIR)$(LOGDIR)/recap
 	@install -dm0755 $(DESTDIR)$(LOGDIR)/recap/backups
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ The cron file (`/etc/cron.d/recap`) is used to determine the execution time of `
 
 ### Configuration
 
-The following variables are commented out with the defaults values in the configuration file `/etc/recap` which can be overriden.
+The following variables are commented out with the defaults values in the configuration file `/etc/recap.conf` which can be overriden.
 
 #### Settings shared by recap scripts
 
--- a/src/recap
+++ b/src/recap
@@ -17,17 +17,6 @@
 #   51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 #
-# The cron execution file is in /etc/cron.d
-#
-# The variable configuration file for recap is in /etc/recap
-#
-# All the variables defined here and several others can be defined in the variable
-# configuration file.  You can see more on these in /usr/share/doc/recap-{version}/README
-#
-# Define the default environment for recap
-# DO NOT change these values here, user-modifiable options can be tweaked through
-# /etc/recap
-#
 #~ Usage: _tool_ [OPTION]
 #~ Options:
 #~   -h,--help       Print this help.
@@ -697,6 +686,29 @@ check_disk_space() {
 }
 
 ## Main
+
+# Check for the configuration file.
+# The following create awareness about the new the config location in
+# /etc/recap.conf, taking precendence the actual /etc/recap config.
+if [[ -r /etc/recap &&
+      -r /etc/recap.conf ]]; then
+  log WARNING "Configuration files exist at old(/etc/recap) and"\
+              "new(/etc/recap.conf) locations. The file from the old"\
+              "location will be read."
+  log WARNING "Please move your configuration to /etc/recap.conf."
+  source /etc/recap
+elif [[ -r /etc/recap &&
+        ! -r /etc/recap.conf ]]; then
+  log WARNING "Configuration file exists at old(/etc/recap) location. "\
+              "The file will be read."
+              "Please move your configuration file to /etc/recap.conf."
+  source /etc/recap
+elif [[ ! -r /etc/recap.conf ]]; then
+  log WARNING "No configuration file found. Expecting /etc/recap.conf."
+  log WARNING "Proceeding with defaults."
+else
+  source /etc/recap.conf
+fi
 
 # Evaluate input flags, print usage if more than one flag is used.
 if [[ "$#" -gt 1 ]]; then
--- a/src/recap.5
+++ b/src/recap.5
@@ -28,7 +28,7 @@ recap - dumps periodic information about running applications and resource usage
 recap is a user-configurable script that can be run once, or run periodically out of cron to dump information about running processes and resource usage. It is useful on servers that have periodic, mysterious performance anomalies for tracking down what may be going on at the time of any particular incident.
 
 The default values for which reports are generated and how many reports are stored can be
-.IR /etc/recap "."
+.IR /etc/recap.conf "."
 The output files from the script are written to
 .IR BASEDIR "."
 .SH OPTIONS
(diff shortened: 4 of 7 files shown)


troyengel:
👍 - I very much like this change because ```/etc/sysconfig/``` (I didn't know that was in there) is a Red Hat specific invention and not part of the [FHS spec](http://www.pathname.com/fhs/pub/fhs-2.3.html#ETCHOSTSPECIFICSYSTEMCONFIGURATION). This update gets rid of that vendor specific code and makes it distro agnostic.

carlwgeorge:
@troyengel This doesn't have anything to do with `/etc/sysconfig`, despite the name similarity to the Makefile variable `$(SYSCONFDIR)` (or the RPM macro `%{_sysconfdir}` for that matter).  We're talking about just plain `/etc` here.

@tonyskapunk The only problem I see with a check for the old location `/etc/recap` is that on RPM upgrades, the file would be renamed to `/etc/recap.rpmsave`.  What do you think about having the code check for that location as well?

troyengel:
@carlwgeorge - you may have missed it in the diff, I was referring to the code where we are, in fact, removing the ```/etc/sysconfig``` check (I imagine this is from a long time) right here: https://github.com/rackerlabs/recap/pull/137/commits/ed240363cdd9318d059eb679e9e52c2836bd3d0a#diff-cf77cb5672a68c31afd4bc3025d4288cL611

On the .rpmsave -- won't this happen only if the configuration file has not been edited by the end user? (matches the hash from the RPM) I'm not too familiar with the semantics when the name *changes* and it's been edited though at the same time.  If we build in a check for .rpmsave (or maybe ,rpmnew?), should we also add the same check for whatever DEB packages do in the same conceptual ballpark? Cover all our bases?

* http://people.ds.cam.ac.uk/jw35/docs/rpm_config.html

If we have %config(noreplace) I think (if I read the chart correctly for our situation) if the user has edited the file the new one will become recap.conf.rpmnew? (http://pkgs.fedoraproject.org/cgit/rpms/recap.git/tree/recap.spec#n50)

carlwgeorge:
Ah I see the older check now, that's what I get for skimming.

Your edit hits the nail on the head, they are two separate files.  Your point about DEB semantics is noted, and actually changes my mind about having recap do any package-specific things.

What about having the RPM own both files?

Brainstorm: https://gist.github.com/carlwgeorge/eed7aaeadc4f30e729af33416be24170

tonyskapunk:
I want to avoid doing something distro specific in the code, hence reading `.rpmnew` and others ideally should be avoided.

The purpose of this PR is:
- Introduce the new location `/etc/recap.conf`
- Create awareness to users of the new location under the following conditions:
  - Old and New locations exist --> Old takes precedence and a message is displayed[0].
  - Only Old location exists --> Config is loaded and a message is displayed[1].
  - Both locations are not existent --> Defaults are used and a message is displayed[2].

Now back to the package discussion.

1. Tracking two locations `/etc/recap` and `/etc/recap.conf`
   1. unedited:
        - The code is able to support this scenario, the old location takes precedence for now.
        - The code should not modify the old config, despite being the same.
   1. edited:
        - The code is able to support this scenario, the old location takes precedence for now.
        - The code should not modify the rpm saved config.

1. Tracking just the new one `/etc/recap.conf`
  In this case the `/etc/recap` will be around but it really does not matter as the code is able to read from this file as it takes precedence for now, but still will notify the user there is a new version in place.

I think tracking just one file is less effort and has almost the same result.

Thoughts?

---

[0] 
  ```bash
Configuration files exist at old(/etc/recap) and new(/etc/recap.conf) locations. The file from the old location will be read. 
 Please move your configuration to /etc/recap.conf.
```
[1]
  ```bash
Configuration file exists at old(/etc/recap) location.  The file will be read. 
 Please move your configuration file to /etc/recap.conf.
```
[2]
  ```bash
No configuration file found. Expecting /etc/recap.conf
 Proceeding with defaults.
```

**Comment by carlwgeorge** [comment]
Allow me to clarify what will happen if we simply do this in the spec file:

```patch
-%config(noreplace) %{_sysconfdir}/recap
+%config(noreplace) %{_sysconfdir}/recap.conf
```

### unedited config

* everything happens as expected, no user interaction

### edited config

* `/etc/recap` will be renamed to `/etc/recap.rpmsave`
* new default config will be at `/etc/recap.conf`
* since `/etc/recap` does not exist, the program ignores all previous customizations

I'd like to avoid that last scenario to best comply with [EPEL updates policy](https://fedoraproject.org/wiki/EPEL_Updates_Policy).

tonyskapunk:
Understood, then the best approach initially is to keep track of them both.

Now, for future releases how it is handled to get rid of the old `/etc/recap` ? 

I want to know if there is anything the code should take into consideration but at the same time I want to avoid being too distro specific as reading  `.rpm{new,save}` as my intentions in the following releases are:
- `1.4.0`: 
  - Create awareness of `/etc/recap` but now make `/etc/recap.conf` to take precedence.
  - If `/etc/recap.conf` is missing but `/etc/recap` is found the latter is read.
- `1.5.0`:
  - Ignore `/etc/recap`, `/etc/recap.conf` is read otherwise defaults are used.

tonyskapunk:
Oh, just realized that forgot to tag @carlwgeorge in my previous comment, any thoughts on the path of deprecation?  If all looks good I'll go ahead and merge.  Thanks!

carlwgeorge:
I agree about avoiding reading `.rpmsave/.rpmnew` files.  I need some more time to think through the best way to handle the migration on the RPM side.  Specifically I want to see how the rpmsave/rpmnew mechanism would work if `/etc/recap` is a symlink to `/etc/recap.conf`.

tonyskapunk:
Hi @carlwgeorge sorry for the pressure with this one, since I'm looking into releasing [1.3.0](https://github.com/rackerlabs/recap/milestone/4) next week I'll like to know how are we feeling about this change and if I should delay it one more release to give you more time solving the packaging situation.

Thanks!

tonyskapunk:
Rebased to include the now support of logging in `recap`

carlwgeorge:
After tinkering with this a bit, I think our best option is to implement the rpmsave file handline via a `%posttrans` scriptlet.

```
%posttrans
if [ -f /etc/recap.rpmsave ]; then
    mv -v /etc/recap.conf /etc/recap.conf.rpmnew
    mv -v /etc/recap.rpmsave /etc/recap.conf
fi
```

Here is what would happen during the transaction:

1. recap-$NEW with `/etc/recap.conf` would be installed.
2. recap-$OLD with `/etc/recap` would be uninstalled.
3. If `/etc/recap` was customized, it will be moved to `/etc/recap.rpmsave`.
4. The `%posttrans` script is run, and if it detects `/etc/recap.rpmsave`, it moves `/etc/recap.conf` out of the way and puts the customized config file in it's place.

The actual recap script can handle the transition however you like, I'll just add that to the spec file as soon as reading `/etc/recap.conf` is supported.

troyengel:
👍 - I like it. Is there any concern if - somehow, not sure how - an existing `/etc/recap.conf.rpmnew` is already on disk? It would get clobbered, but I think in RPM terminology even if that did happen the `.rpmnew` file means "I'm the unedited version from an RPM" so there's no altered content in there, yah?

carlwgeorge:
> the .rpmnew file means "I'm the unedited version from an RPM" so there's no altered content in there, yah?

:+1: 

No different than if subsequent package updates had new config files.  For example if `-2` and `-3` each had a different config file from `-1`, and you updated through each one, `-3`'s `.rpmnew` file would overwrite `-2`'s `.rpmnew` file.

Good point on the force flag, I'll make sure to use that in the actual spec file.

tonyskapunk:
To `recap`-itulate  ;) ...

- If the old `/etc/recap` config is unedited it will be uninstalled and the new config `/etc/recap.conf` will be used.
- If the old `/etc/recap` config is edited, it will be moved to `/etc/recap.conf` to remain in use, the new `/etc/recap.conf` will be moved to `/etc/recap.conf.rpmnew`.

Subsequent releases will behave exactly the same, leaving the edited config file in `/etc/recap.conf` and the config file from the package in `/etc/recap.conf.rpmnew`.

It will be up to the user to ensure they switch to the latest configuration.

tonyskapunk:
Merging this one into `development`

PR #143, opened 2017-12-06
Release 1.3.0

tonyskapunk:
- Start the deprecation of `/etc/recap` config file for `/etc/recap.conf`.
- Add systemd support for units and timers.
- Add log capabilities for recap.
- Bug fixing (logging and clean up functions).

PR #145, opened 2017-12-07
Makefile tweak

carlwgeorge:
While looking over #143 I noticed a few inefficiencies.

* `ls` child process for a file listing
* `echo` and `sed` child processes to trim a few characters from a variable name


tonyskapunk:
👍 Merging

PR #150, opened 2018-04-13
Remove SIGSTOP(19), Fix #149.

### LINKED ISSUE #149: remove 19)SIGSTOP from trap
Sources:
https://github.com/rackerlabs/recap/blob/master/src/recap#L182
https://github.com/rackerlabs/recap/blob/master/src/recaplog#L220

tonyskapunk:
```bash
$ man trap | grep SIGSTOP
       extension. Setting a trap for SIGKILL or SIGSTOP produces undefined results.
       Trapping  SIGKILL  or  SIGSTOP is syntactically accepted by some historical implementations, but it has no effect. Portable POSIX

```

### DIFF
--- a/src/recap
+++ b/src/recap
@@ -179,7 +179,7 @@ recaplock() {
     log ERROR "$( basename $0 )[$$]: Lock File exists - exiting"
     exit 1
   else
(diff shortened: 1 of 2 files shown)


tonyskapunk:
Merging now.

PR #154, opened 2018-05-04
Add support to multiple mysql config files.

### LINKED ISSUE #136: Add support for multiple mysql instances running of different ports.
Please add support to monitor multiple mysql instances running on different ports.

tonyskapunk:
- Adding support to multiple mysql config files
- Before using a config file it ensures the file is avalable
- Errors when reading such files will be logged into recap.log

Fix #136

### DIFF
--- a/src/recap
+++ b/src/recap
@@ -351,13 +351,14 @@ print_netstat_sum() {
 print_mysql() {
   log INFO "Starting 'mysql status' report"
   local LOGFILE="$1"
+  local MYCNF="$2"
   local PLESK_FILE="/etc/psa/.psa.shadow"
   if [[ -r "${PLESK_FILE}" ]]; then
-    echo "MySQL status" >> "${LOGFILE}"
+    echo "MySQL (plesk) status" >> "${LOGFILE}"
     mysqladmin -uadmin -p$(cat "${PLESK_FILE}") status >> "${LOGFILE}"
   else
-    echo "MySQL ${DOTMYDOTCNF} status" >> "${LOGFILE}"
-    mysqladmin --defaults-extra-file="${DOTMYDOTCNF}" status >> "${LOGFILE}"
+    echo "MySQL (${MYCNF}) status" >> "${LOGFILE}"
+    mysqladmin --defaults-extra-file="${MYCNF}" status >> "${LOGFILE}"
   fi
   print_blankline "${LOGFILE}"
   log INFO "Ended 'mysql status' report"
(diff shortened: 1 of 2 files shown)


tonyskapunk:
Adding an example on  #136

jamrok:
Hey,

Sorry for the delay in testing but I got to test this yesterday and found a bug and submitted a [PR against your branch](https://github.com/tonyskapunk/recap/pull/3).

If the default mysql option file `/root/.my.cnf` exists but the instance that it references isn't accessible (eg. stopped, crashed, etc), recap won't record the stats for any other MySQL instance (Scenario A) and depending on your options, recap will hang (Scenario B) and block subsequent recap runs due to its lock file not being released.

**Scenario A** shows the following output in the mysql recap log (this does not hang):
```
MySQL (/root/.my.cnf2) status
 
MySQL (/root/.my.cnf2) processes
```
Relevant `/etc/recap.conf` config:
```
USEMYSQL="yes"
USEMYSQLPROCESSLIST="yes"
USEINNODB="no"
DOTMYDOTCNF="/root/.my.cnf1,/root/.my.cnf2"
```

**Scenario B** shows the following output in the mysql recap log (this blocks future recap runs):
```
MySQL (/root/.my.cnf2) status

MySQL (/root/.my.cnf2) processes
 
MySQL (/root/.my.cnf2) InnoDB status
```

Relevant `/etc/recap.conf` config:
```
USEMYSQL="yes"
USEMYSQLPROCESSLIST="yes"
USEINNODB="yes"
DOTMYDOTCNF="/root/.my.cnf1,/root/.my.cnf2"
```

The reason for this is due to using the `--defaults-extra-file` option instead of the `--defaults-file` option. The [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/option-file-options.html#option_general_defaults-extra-file) state:
>--defaults-file=file_name
>Read only the given option file. If the file does not exist or is otherwise inaccessible, an error occurs. 

Since the mysql* binaries looks for the `/root/.my.cnf` file by default, when it's found, it's read, and if mysql is inaccessible, it returns an error and then exits and causes the issues mentioned above.

Is there any reason why we need to use `--defaults-extra-file`? If not, the PR should be good. The hanging (`cat` executed without parameters) in Scenario B occurred due to not checking that we had valid data returned from the query, hence the variable validation in the PR. This shouldn't be an issue if we switch to `--defaults-file`, but added the validation nonetheless.

Outside of that everything else works as expected, eg. whether or not the mysql option file was specified in `DOTMYDOTCNF` but don't actually exist or the DB instance is inaccessible, toggling options in different states (on/off), etc.

tonyskapunk:
Thanks for the PR, conversation regarding this can be found here https://github.com/tonyskapunk/recap/pull/3#issuecomment-404299392

tonyskapunk:
With the points covered I'm finally merging this one.

PR #157, opened 2018-05-11
Setting right permissions to BASEDIR on runtime.

### LINKED ISSUE #156: recap recreates log dir(BASEDIR) with loose permissions
The `Makefile` creates the directory with the correct permissions: `0750`

https://github.com/rackerlabs/recap/blob/master/Makefile#L77

But if the log dir(`BASEDIR`) is not in place whenever `recap` runs it will create the directory to place the logs in there, the problem is the permissions in there rely on umask,(default in most linux distros: `0022`):

`BASEDIR` default: https://github.com/rackerlabs/recap/blob/master/src/recap#L45
Creation of the `BASEDIR` if it does not exist on runtime: https://github.com/rackerlabs/recap/blob/master/src/recap#L198

tonyskapunk:
Fix #156

### DIFF
--- a/src/recap
+++ b/src/recap
@@ -198,6 +198,7 @@ create_output_dirs() {
       mkdir -p "${OUTPUT_DIR}"
     fi
   done
+  chmod 0750 "${BASEDIR}"
 }
 
 # Create output file


tonyskapunk:
Merging, this is an quick win

PR #160, opened 2018-06-08
Adding explicit commands to (en|dis)able systemd timers.

### LINKED ISSUE #159: Activate systemd timers via make
When installing via make (eg. on Ubuntu systems), the cron is automatically enabled when `systemd` is not present - yet, the timer is _not_ activated when it is.

I understand there is a notice about such when installing, but the behaviour is not consistent or expected for people familiar with how `recap` normally works via package (or when not using `systemd`).

I would propose that as part of the make we enable the appropriate timers, or if we want to leave it up to the end user, give them some copy/paste commands. I had someone today not familiar about systemd timers get confused about the lack of cron job and went and added it manually.

tonyskapunk:
- Make install/uninstall will include commands to enable/disable the timers.

Fix #159


tonyskapunk:
Most of the comments regarding this were in the issue #159 I'm going to merge this one in development, this PR does not perform changes in the behavior just on the print out information when installing timers.

PR #167, opened 2018-10-06
Use systemd ansible module for timers.

tonyskapunk:
- Use `systemd` ansible module instead of shell
- Enables compatibility with debian 8 (Jessie)


tonyskapunk:
Last commit ensures multiple runs (installs or uninstalls) succeed despite of previous conditions.

PR #168, opened 2018-10-19
Add base travis config.

tonyskapunk:
- Add initial travis config.


tonyskapunk:
This is building successfully, and will work as a baseline for the tests and builds to come.

PR #171, opened 2018-11-09
Make http_status (USEFULLSTATUS) a plugin

### LINKED ISSUE #169: Links package requirement - AWS Linux
In AWS Linux, Links is a very hard package to get installed. You quickly hit some dependency mayhem that just takes too long to resolve, when this feature is normally disabled by default. My solution would be to remove this feature from the source, and release a version that can be installed through the epel repository. 

Here is what worked for me:

wget https://github.com/rackerlabs/recap/archive/master.zip
unzip master.zip
cd recap-master/
make install

Thanks!

Chris Mendoza

### LINKED ISSUE #170: Replace 'links' with 'elinks'
While adding recap to a cloud server today (default Linode.com CentOS 7 image), the dependency chain of `links` drags in a ton of stuff we do not need, a lot of X graphical things.  Conversely, the `elinks` version (which I converted to years ago on my workstations) only has 2 minimal packages as deps. The `elinks` binary supports `-dump` as well, so should be a painless transition.

Supporting output - first, the yum install links list:

```
# yum install links
...
 Package                Arch        Version                     Repository
                                                                           Size
================================================================================
Installing:
 links                  x86_64      1:2.13-1.el7                epel      2.8 M
Installing for dependencies:
 cairo                  x86_64      1.14.8-2.el7                base      713 k
 fontconfig             x86_64      2.10.95-11.el7              base      229 k
 gdk-pixbuf2            x86_64      2.36.5-1.el7                base      567 k
 graphite2              x86_64      1.3.10-1.el7_3              base      115 k
 harfbuzz               x86_64      1.3.2-1.el7                 base      177 k
 jasper-libs            x86_64      1.900.1-31.el7              base      150 k
 jbigkit-libs           x86_64      2.0-11.el7                  base       46 k
 libXdamage             x86_64      1.1.4-4.1.el7               base       20 k
 libXext                x86_64      1.3.3-3.el7                 base       39 k
 libXfixes              x86_64      5.0.3-1.el7                 base       18 k
 libXft                 x86_64      2.3.2-2.el7                 base       58 k
 libXrender             x86_64      0.9.10-1.el7                base       26 k
 libXxf86vm             x86_64      1.1.4-1.el7                 base       18 k
 librsvg2               x86_64      2.40.16-1.el7               base      128 k
 libthai                x86_64      0.1.14-9.el7                base      187 k
 libtiff                x86_64      4.0.3-27.el7_3              base      170 k
 libwayland-client      x86_64      1.14.0-2.el7                base       32 k
 libwayland-server      x86_64      1.14.0-2.el7                base       38 k
 libxshmfence           x86_64      1.2-1.el7                   base      7.2 k
 mesa-libEGL            x86_64      17.2.3-8.20171019.el7       base       96 k
 mesa-libGL             x86_64      17.2.3-8.20171019.el7       base      156 k
 mesa-libgbm            x86_64      17.2.3-8.20171019.el7       base       38 k
 mesa-libglapi          x86_64      17.2.3-8.20171019.el7       base       43 k
 pango                  x86_64      1.40.4-1.el7                base      275 k
 pixman                 x86_64      0.34.0-1.el7                base      248 k

Transaction Summary
================================================================================
Install  1 Package (+25 Dependent packages)
```

The yum install elinks:

```
# yum install elinks
...
================================================================================
 Package               Arch         Version                    Repository  Size
================================================================================
Installing:
 elinks                x86_64       0.12-0.37.pre6.el7         base       882 k
Installing for dependencies:
 js                    x86_64       1:1.8.5-20.el7             base       2.3 M
 nss_compat_ossl       x86_64       0.9.6-8.el7                base        37 k

Transaction Summary
================================================================================
Install  1 Package (+2 Dependent packages)
```

So while attempting to add recap to just capture my statistics (not even using this links feature at all), I'm faced with adding 20+ extra meaningless packages to a minimal footprint cloud device. We can reduce this cruft by switching to `elinks` and lose no functionality, since all that recap does is run `elinks -dump (url)` and prints it, no parsing.  The `elinks` output is very links-similar.

Additionally, notice that `elinks` comes from Base, it's the standard in modern Red Hat. `links` has been pushed to EPEL as it was removed from RHEL long ago, so we're working closer towards removing an external (non-vendor) dependency chain.

Related Issue: #169  - this may help fix that problem

tonyskapunk:
- Move http_status report out of `resources` report.
- Adjust  `recaplog` to find logs of the form `name_(subname_)*_<date>`.
- Remove `USEFULLSTATUS` from `recap.conf`
- Replace **links** with **elinks** to reduce dependencies.

Important notes about the functional changes:
- Prior to this change  `USEFULLSTATUS` now `http_status` was disabled by default.
  - Now `http_status` is a plugin, hence is disabled by default.
- The reports of `http_status` were generated in the `resources` report, now generates its own report as any other plugin.
- Configuration examples in `recap.conf` have been removed, any modification to the default configurations for this `http_status` plugin will need to be added, this is true for any other plugin.

Notes on upgrade and versioning:
- This change is targeted for a newer major version of `recap` (2.0.0), per [Versioning](https://github.com/rackerlabs/recap#versioning) notes, this is a major change that will not be backward compatible.
- When upgrading, if the previous configuration had `USEFULLSTATUS` enabled, the behaviour will change since:
  1. The enable plugin option in `recap.conf` is disabled by default
  1. The plugin to generate reports needs to be enabled.
  This implies that it will be required to enable plugins and the `http_status` plugin itself. 
- Similarly if there is any configuration change to the previous `USEFULLSTATUS` options those will need to be re-added to the `recap.conf` with the new valid names

---
Fix #170, Fix #169

### DIFF
--- a/src/recaplog
+++ b/src/recaplog
@@ -89,7 +89,7 @@ delete_old_logs() {
   matched_files=$(
     find "${BASEDIR}" -maxdepth 2 -regextype posix-extended \
       -type f -mtime +${LOG_EXPIRY} -regex \
-      "^${BASEDIR}/(.*/)?[a-z]+_([a-z]+_)?[0-9]{8}(-[0-9]{6})?\.log(\.tar)?(\.gz)?$"
+      "^${BASEDIR}/(.*/)?[a-z]+_([a-z]+_)*[0-9]{8}(-[0-9]{6})?\.log(\.tar)?(\.gz)?$"
   )
   if [[ -n "${matched_files}" ]]; then
     count_files=$( wc -l <<<"${matched_files}" )
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Contribution guidelines can be found in [CONTRIBUTING.md](https://github.com/rac
 - grep
 - iotop
 - iproute/iproute2
-- links
+- elinks
 - procps
 - psmisc
 - sysstat >= 9
@@ -337,7 +337,7 @@ These are the type of reports generated and their dependencies.
 
 - **USERESOURCES** - Generates "resources"(uptime, free, vmstat, iostat, iotop) log
 
-  Required by: `USEDF`, `USESLAB`, `USESAR`, `USESARQ`, `USESARR`, `USEFULLSTATUS`
+  Required by: `USEDF`, `USESLAB`, `USESAR`, `USESARQ`, `USESARR`
 
   Default: `USERESOURCES="yes"`
 
@@ -375,12 +375,6 @@ These are the type of reports generated and their dependencies.
 
   Default: `USESARR="no"`
 
-- **USEFULLSTATUS** - Performs an http request(GET) to the URL defined in `OPTS_STATUSURL`. Needs a webserver configured to respond to this request. Nginx(nginx_status) and Apache HTTPD(server-stats) offer this functionality that needs to be enabled.
-
-  Requires: `USERESOURCES`
-
-  Default: `USEFULLSTATUS="no"`
-
 
 #### Options
 
@@ -397,11 +391,6 @@ Options used by the tools generating the reports
   Required by: `USEMYSQLPROCESSLIST`
 
   Default: `MYSQL_PROCESS_LIST="table"`
-
-- **OPTS_LINKS** - Options used by links.
-  Required by: `USEFULLSTATUS`
-
-  Default: `OPTS_LINKS="-dump"`
 
 - **OPTS_DF** - df options
 
@@ -457,12 +446,6 @@ Options used by the tools generating the reports
 
   Default: `OPTS_PSTREE="-p"`
 
-
-- **OPTS_STATUSURL** - URL to perform the http request  when USEFULLSTATUS is enabled.
-
-  Required by: `USEFULLSTATUS`
-
-  Default: `OPTS_STATUSURL="http://localhost:80/server-status"`
 
 - **OPTS_VMSTAT** - vmstat options
 
--- a/src/core/resources
+++ b/src/core/resources
@@ -83,15 +83,6 @@ print_sar() {
   log INFO "Ended 'sar' report"
 }
 
-# Print the output of the web server status page to the specified file
-print_http_fullstatus() {
-  local LOGFILE="$1"
-  log INFO "Starting 'web status page' report - ${LOGFILE##*/}"
-  echo "Web Status report" >> "${LOGFILE}"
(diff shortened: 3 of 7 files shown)


src/recaplog
```diff
@@ -89,7 +89,7 @@ delete_old_logs() {
   matched_files=$(
     find "${BASEDIR}" -maxdepth 2 -regextype posix-extended \
       -type f -mtime +${LOG_EXPIRY} -regex \
-      "^${BASEDIR}/(.*/)?[a-z]+_([a-z]+_)?[0-9]{8}(-[0-9]{6})?\.log(\.tar)?(\.gz)?$"
+      "^${BASEDIR}/(.*/)?[a-z]+_([a-z]+_)*[0-9]{8}(-[0-9]{6})?\.log(\.tar)?(\.gz)?$"
```
terackspace: Can you explain what this update to the regex is doing (in real life)? We're going from a single character match to a greedy wildcard, it may match more that expected? (meaning does it need to be non-greedy, like `.*?` to ensure the RHS of it doesn't get sucked into the greed?)

tonyskapunk: The important change is
From:
  ```
  ([a-z]+_)?
  ```
To:
  ```
 ([a-z]+_)*
  ```

In the past we did only have logs as these examples:
- `ps_20181112-111200.log`
- `ps_daily_20181028.log.tar.gz`

With `http_status` and any  other plugin using more than one single word the format will be (examples):
- `http_status_20181112-111200.log`
- `http_status_daily_20181112-111200.log.tar.gz`

Hence the change. Hope that helps to clarify that change.

terackspace: gotcha! I fired up <https://www.debuggex.com/> and played a bit with random input, matches expectations. +1

tonyskapunk:
Rebased to make use of the new tests in #173

terackspace:
The builds are failing as the CI/CD does not have a webserver listening on port 80; properly testing ON and OFF is the realm of issue #178. Our difficulty herein is that the URL is provided as a full object in one variable, making it hard to "tickle" first to see if it's even up.

The only "easy" way to avoid this is set up the code to skip the check if an envvar is set, then when using Travis (or any CI/CD) we could export that for the code to use (if `CI_BUILD==1`detected, it sidesteps the URL check). It's not the most graceful but it would solve the Travis build failures, at the expense of a weird bit of code logic. Maybe using a generic envvar like `SKIP_HTTPSTATUS=1`, that way if a user needed it for some reason...? _shrug_

PR #173, opened 2018-11-16
Add multi-distro testing on travis.

tonyskapunk:
- Adding tests for travis
- Use multiple distros for the testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,31 +7,50 @@ branches:
     - development
     - testing
 
-addons:
-  apt:
-    update: true
-    packages:
-      - bash
-      - coreutils
-      - gawk
-      - grep
-      - iotop
-      - iproute2
-      - elinks
-      - procps
-      - psmisc
-      - sysstat
+matrix:
+  include:
+    - name: "CentOS 6"
+      env:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -7,31 +7,50 @@ branches:
...
+      --rm \
+      --tty \
+      --name recap_on_${DISTRO/:/_} \
+      --mount type=bind,src="$(pwd)",dst=/recap \
+      ${DISTRO}
+  - docker exec recap_on_${DISTRO/:/_} /recap/tests/install_deps.sh ${DISTRO}
+
+script:
+  - docker exec recap_on_${DISTRO/:/_} /recap/tests/test_install_recap.sh
+  - docker exec recap_on_${DISTRO/:/_} recap
```
jamrok: In the build this succeeds, but for testing it should [fail here]( https://travis-ci.org/rackerlabs/recap/jobs/456137856#L629-L630 ) if any error occurs.

tonyskapunk: Maybe, what I'm thinking is 
1. Figure out the reason why iotop fails (I think this is related to the environment, container)
2. Add tests to validate each report(?), plugin(?)

PR #184, opened 2018-12-26
List plugins instead of expand them.

### LINKED ISSUE #183: Error when enabling plugins but none is available.
There is a bug in the code that loads the code of the plugins, when ` USEPLUGINS="yes"` is enabled, the [globbing expansion](https://github.com/rackerlabs/recap/blob/development/src/recap#L255) is not working as expected when there is no content in the `plugins-enabled` directory.

```log
+ echo '2018-12-26 11:45:46-06:00 [INFO] Loading plugin: /usr/local/lib/recap/plugins-enabled/*'
+ source '/usr/local/lib/recap/plugins-enabled/*'
```

tonyskapunk:
This fixes #183  Instead of expanding through globbing it lists the content of the `plugins_dir`, displaying the proper warning when there's none.

```log
2018-12-26 12:06:03-06:00 [WARNING] No plugins found to execute in: /usr/local/lib/recap/plugins-enabled
```

### DIFF
--- a/src/recap
+++ b/src/recap
@@ -252,7 +252,7 @@ plugins_info(){
 plugins_load() {
   local plugin_dir="$1"
   log INFO "Loading plugins from: ${plugin_dir}"
-  for plugin in "${plugin_dir}"/*; do
+  for plugin in $( ls -1 "${plugin_dir}" ); do
     log INFO "Loading plugin: ${plugin}"
     source ${plugin}
   done
--- a/src/recap
+++ b/src/recap
@@ -252,10 +252,10 @@ plugins_info(){
 plugins_load() {
   local plugin_dir="$1"
   log INFO "Loading plugins from: ${plugin_dir}"
-  for plugin in $( ls -1 "${plugin_dir}" ); do
+  while read plugin; do
     log INFO "Loading plugin: ${plugin}"
-    source ${plugin}
-  done
+    source "${plugin_dir}/${plugin}"
+  done < <(ls -1 "${plugin_dir}" 2>/dev/null)
 }
 
 # Load core functions


src/recap
```diff
@@ -252,7 +252,7 @@ plugins_info(){
 plugins_load() {
   local plugin_dir="$1"
   log INFO "Loading plugins from: ${plugin_dir}"
-  for plugin in "${plugin_dir}"/*; do
+  for plugin in $( ls -1 "${plugin_dir}" ); do
     log INFO "Loading plugin: ${plugin}"
     source ${plugin}
```
jamrok: It seems the `plugin_dir` is missing from the sourced file and ideally it should be quoted:
`source "${plugin_dir%/}/${plugin}"`
or the following if you want to get rid of all the possible extra slashes _(I know they don't hurt)_ in the path (eg: `/tmp/plugin//file`):
```
plugin_file=$(realpath -m "${plugin_dir}/${plugin}")
source "${plugin_file}"
```

src/recap
```diff
@@ -252,7 +252,7 @@ plugins_info(){
 plugins_load() {
   local plugin_dir="$1"
   log INFO "Loading plugins from: ${plugin_dir}"
-  for plugin in "${plugin_dir}"/*; do
+  for plugin in $( ls -1 "${plugin_dir}" ); do
```
jamrok: This is what I used to do: use `for` loops for going over files, but if the files have spaces in the name, then that becomes an issue. My solution was to temporarily reset the `IFS` to a new line, but I didn't like that.

For most use cases I started doing this:
```
command 2>/dev/null | while read var; do
  echo "${var}";
done
```
This reads the entire line including spaces instead of splitting on spaces as the `for` loop does due to the IFS. The only issue I've had with this is not being able to use variables outside the loop that were declared inside it (due to the loop being in a subshell).

The way I do it now, which seems best is:
```
while read var; do
  echo "${var}"; 
done < <(command 2>/dev/null)
```
Why? The loop isn't in a subshell so I can use variables without a problem; and I either get data to work with or not (with no errors).

I'd change the `for` loop to a `while` loop for this reason.

tonyskapunk:
With my changes now it is doing it right:
```
+ source /usr/lib/recap/plugins-enabled/kernel_cmd
++ echo kernel_cmd
```

jamrok:
I've confirmed the same on my end:
```
++ ls -1 /usr/lib/recap/plugins-enabled
+ source /usr/lib/recap/plugins-enabled/docker_top
+ read plugin
+ source /usr/lib/recap/plugins-enabled/http_status
+ read plugin
+ source /usr/lib/recap/plugins-enabled/kernel_cmd
+ read plugin
+ source /usr/lib/recap/plugins-enabled/redis
+ read plugin
```
The build shows more info as well:
Before: https://travis-ci.org/rackerlabs/recap/jobs/472410633#L767-L772
After: https://travis-ci.org/rackerlabs/recap/jobs/475453824#L793-L804

tonyskapunk:
Great! merging now

PR #190, opened 2019-01-22
Add debuging and tracing to install test.

tonyskapunk:
- Enable debug and trace on install test
- Enable debug on exec test.


tonyskapunk:
This was meant for testing, an issue is being created as the problem is with bash 5 and trap.

PR #192, opened 2019-01-24
Fixes change in SIGCHLD behaviour in bash 5.0

### LINKED ISSUE #191: Need bash 5.0 support.
Build on bash 5 is currently failing: https://travis-ci.org/rackerlabs/recap/jobs/483067424 reflected only in `master`,

`dev` branch passed fine because it was running bash version 4 (https://travis-ci.org/rackerlabs/recap/jobs/477402022#L1057)  but since bash 5.0 was released lately (https://lists.gnu.org/archive/html/bash-announce/2019-01/msg00000.html), buster picked it up recently (https://packages.debian.org/source/testing/bash) and new runs on it are failing.

jamrok:
Fixes #191 

After investigating this with @tonyskapunk, we found the following:

From https://tiswww.case.edu/php/chet/bash/NEWS:
> n. The SIGCHLD trap is run once for each exiting child process even if job
   control is not enabled when the shell is in Posix mode.

Or on the unofficial mirror: https://github.com/bminor/bash/blob/master/NEWS#L46-L47

Based on the change above, it seems we're now able to do more granular handling of the exit status of child processes by trapping SIGCHLD in the parent and checking the exit status in a "trap handler" function. Since we don't need that, we can remove 17 from the trap list.


tonyskapunk:
lgtm

:trophy:  Thanks @jamrok !

PR #201, opened 2019-02-15
Add an option to enable plugins via ansible.

tonyskapunk:
- Add a way to enable plugins via Ansible


tonyskapunk:
Merging into `development`

PR #202, opened 2019-02-15
Add system lock plugin. 

tonyskapunk:
- Add system lock plugin. 

Fix #202


tonyskapunk:
Merging into `development`

PR #212, opened 2019-04-12
Add fedora distro for CI

tonyskapunk:
Adding Fedora distros for testing:
- latest (currently 29)
- rawhide (currently 31)


tonyskapunk:
rebasing to include #213

PR #219, opened 2019-05-08
Update iotop default to include -P

### LINKED ISSUE #218: resources iotop only show processes
It might be easier to read the resources file when iotop is enabled if we pass `-P` by default.

You can see the clutter on a server running a daemon with a lot of threads (eg mysqld), where `-P` will show only a single line for all threads of a process (consolidated)

tonyskapunk:
- Updates the iotop default to use report per process instead of the default per thread

Fixes #218


tonyskapunk:
Merging into dev since there is no breaking changes.

PR #224, opened 2019-07-05
Add CLI plugins to recap

tonyskapunk:
- Add a new option `-p` for plugin actions in `recap`
- Lists, Enables or Disables plugins
- Moves out of recap the plugin functions
- Updates to `man`, `CHANGELOG.md` to reflect new option
- Add testing in travis

### DIFF
--- a/src/core/plugins
+++ b/src/core/plugins
@@ -25,9 +25,9 @@ in_array() {
   local element="${1}"
   shift
   local -a array=( ${@} )
-  
+
   # Copy the array, removing the element
-  c_array=( ${array[@]//${element}} )
+  local -a c_array=( ${array[@]//${element}} )
 
   # If the arrays are the same, the element is not part of the array
   if [[ ${#array[@]} == ${#c_array[@]} ]]; then
@@ -37,8 +37,8 @@ in_array() {
   fi
 }
 
-# Removes dups from a csv string
-csv_remove_dups() {
+# Transforms a csv into a space-separated list, also removs dups
+csv_to_list() {
   local csv="${1}"
   local -a array
 
@@ -57,8 +57,8 @@ plugins_list() {
   echo ${p_names[@]}
 }
 
-# Validate action (list, enable, disable)
-plugin_validate_action() {
+# Validate plugin action (list, enable, disable)
+validate_plugin_action() {
   case "${1}" in
     "list"|"enable"|"disable")
       return 0
@@ -69,8 +69,20 @@ plugin_validate_action() {
   esac
 }
 
-# Obtain a list of plugin by status
-get_plugin_status() {
+# Validate plugin status (available, enabled, disabled)
+validate_plugin_status() {
+  case "${1}" in
+    "available"|"enabled"|"disabled")
+      return 0
+    ;;
+    *)
+      return 1
+    ;;
+  esac
+}
+
+# Obtain a list of plugins by status
+get_plugin_by_status() {
   local list_status="${1}"
   local -a plugins
 
@@ -84,49 +96,23 @@ get_plugin_status() {
     "disabled")
       plugins_a=( $( plugins_list "${PLUGIN_A_DIR}" ) )
       plugins_e=( $( plugins_list "${PLUGIN_E_DIR}" ) )
-      # Loop through the available plugins
-      for p_a in ${plugins_a[@]}; do
-        # Copy the list of enabled plugins removing the available in turn
-        c_p_e=( ${plugins_e[@]//${p_a}} )
-        # If both arrays are the same then the looped plugin is disabled
-        if [[ ${#plugins_e[@]} == ${#c_p_e[@]} ]]; then
-            plugins+=( ${p_a} )
-        fi
-      done
+      # Get disabled plugins (available but not enabled)
+      plugins=(
+        $( comm \
+             -23 \
+             <( printf "%s\n" "${plugins_a[@]}" ) \
+             <( printf "%s\n" "${plugins_e[@]}" ) |
+               tr '\n' ' '
+        )
+      )
     ;;
   esac
 
   echo "${plugins[@]}"
 }
 
-# Prints the status of plugins
-print_plugins() {
-  local list_status="${1}"
-  local plugins
-
-  case "${list_status}" in
-    "available"|"enabled"|"disabled")
-      plugins=$( get_plugin_status "${list_status}" )
-    ;;
-    *)
-      echo -e "Error: Invalid list status argument: ${list_status}\n" >&2
-      print_usage
-      return 1
-    ;;
-  esac
-
-  if [[ ${#plugins} == 0 ]]; then
-    log ERROR "No ${list_status} plugins found."
-  else
-    plugins=$( echo "${plugins[@]}" | tr " " "," )
-    echo "Plugins ${list_status}: ${plugins%,}"
-    log INFO "Plugins ${list_status}: ${plugins%,}"
-  fi
-  return 0
-}
-
-# Validate plugin names. "all" is valid (represent all available)
-validate_plugin() {
+# Validate plugin name. "all" is valid (represent all available)
+validate_plugin_name() {
   local plugins="$1"
   local -a plugins_a
 
@@ -134,13 +120,13 @@ validate_plugin() {
   if [[ "${plugins}" == "all" ]]; then
     return 0
   fi
-  
+
   # Getting the list of available plugins
-  plugins_a=( $( get_plugin_status "available" ) )
+  plugins_a=( $( get_plugin_by_status "available" ) )
 
   # Create array with the provided plugins, remove dups
-  local -a p_v=( $( csv_remove_dups "${plugins}" ) )
-  
+  local -a p_v=( $( csv_to_list "${plugins}" ) )
+
   # Validate provided plugin with available
   for p in ${p_v[@]}; do
     if ! in_array ${p} ${plugins_a[@]}; then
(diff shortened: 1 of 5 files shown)


tonyskapunk:
@StafDehat if you have some cycles available, would you mind reviewing this? Thanks!

src/core/plugins
```diff
@@ -0,0 +1,306 @@
...
+  # If the arrays are the same, the element is not part of the array
+  if [[ ${#array[@]} == ${#c_array[@]} ]]; then
+    return 1
+  else
+    return 0
+  fi
+}
+
+# Removes dups from a csv string
+csv_remove_dups() {
```
StafDehat: This actually does more than just de-dupe a CSV.  It also turns the CSV into a space-separated list.

src/core/plugins
```diff
@@ -0,0 +1,306 @@
...
+  echo ${array[@]}
+}
+
+
+## Helper functions for plugin cli actions
+
+# Obtain the list of plugins in a directory
+plugins_list() {
+  local plugin_dir="$1"
+  local p_names=( $( ls -1 "${plugin_dir}" | tr '\n' ' ') )
```
StafDehat, line 56: Heads-up, using space as a delimiter here means that any plugin with a space in its filename, will break this function.  That might be acceptable, but if you want to allow it, you'll likely have to move away from arrays a bit, and rely on newline-separated strings.

src/core/plugins
```diff
@@ -0,0 +1,306 @@
...
+    "available")
+      plugins+=( $( plugins_list "${PLUGIN_A_DIR}" ) )
+    ;;
+    "enabled")
+      plugins+=( $( plugins_list "${PLUGIN_E_DIR}" ) )
+    ;;
+    "disabled")
+      plugins_a=( $( plugins_list "${PLUGIN_A_DIR}" ) )
+      plugins_e=( $( plugins_list "${PLUGIN_E_DIR}" ) )
+      # Loop through the available plugins
```
StafDehat: I'd encourage you to use 'comm' or 'diff' here instead of copying an array within a loop.  I think it'd be more efficient, and to me it seems simpler to read.  In addition, if you move away from whitespace-separated lists in the plugins_list() output and use newline-separated lists instead, it'll already be in the right format:
```
# Show plugins in the Available dir, but not in Enabled:
comm -23 <( plugins_list "${PLUGIN_A_DIR}" ) <( plugins_list "${PLUGIN_E_DIR}" )
```

src/core/plugins
```diff
@@ -0,0 +1,306 @@
...
+      return 0
+    ;;
+    *)
+      return 1
+    ;;
+  esac
+}
+
+# Obtain a list of plugin by status
+get_plugin_status() {
```
StafDehat: We should rename this function - get_plugin_status() sounds to me like it'll take a plugin name as an argument, and tell you what its status is.  Call this get_plugins_by_status() or something, instead.

src/core/plugins
```diff
@@ -0,0 +1,306 @@
...
+        fi
+      done
+    ;;
+  esac
+
+  echo "${plugins[@]}"
+}
+
+# Prints the status of plugins
+print_plugins() {
```
StafDehat: This function name should be tweaked too - it suggests it'll print everything, when in actually it takes a status as argument, and prints only the plugins in that status.  Which... sounds a heck of a lot like the function right above this one.  These two functions could likely be simplified into one.

src/core/plugins
```diff
@@ -0,0 +1,306 @@
...
+
+  return 0
+}
+
+# Enable Plugin support in recap.conf
+enable_plugin_conf() {
+   local recap_conf="/etc/recap.conf"
+
+   # Only perform changes when not already defined
+   if [[ "${USEPLUGINS}" == "no" ]]; then
```
StafDehat: To reduce nesting, test for USEPLUGINS=yes, and return.  Then the rest of this function can be 1-less-nest.

src/core/plugins
```diff
@@ -0,0 +1,306 @@
...
+    if ! in_array ${p} ${plugins_a[@]}; then
+        return 1
+    fi
+  done
+
+  return 0
+}
+
+# Enable Plugin support in recap.conf
+enable_plugin_conf() {
```
StafDehat, line 141: Seems like the goal of this whole function is to ensure that recap.conf contains exactly 1 line setting USEPLUGINS, and that it sets it to "yes".  As such, this function seems excessively complicated.
I propose this instead:
```
# If already enabled, just return:
[[ "${USEPLUGINS}" == "yes" ]] && return

# If config already mentions USEPLUGINS, ensure they're all commented:
if grep -qP '^\s*(#.*)?\s*USEPLUGINS\s*=' "${recap_conf}"; then
  # Comment all of 'em:
  sed -i '/^\s*USEPLUGINS\s*=/s/^/#/' "${recap_conf}"
fi

# Add one entry at the bottom to enable USEPLUGINS:
echo 'USEPLUGINS="yes"' >> "${recap_conf}"
```

src/core/plugins
```diff
@@ -0,0 +1,306 @@
...
+  else
+    p_2a=( $( csv_remove_dups "${plugins}" ) )
+  fi
+
+  # Enable or disable plugins that are currently disabled or enabled
+  for plugin in ${p_2a[@]}; do
+    if in_array ${plugin} ${p_a[@]}; then
+      local symlink="${PLUGIN_E_DIR}/${plugin}"
+
+      # Enable disabled plugin
```
StafDehat, line 205: I think we could increase readability by abstracting the actual filesystem operations within "enable_plugin" and "disable_plugin" helper functions.

src/core/plugins
```diff
@@ -0,0 +1,306 @@
...
+        {
+          rm "${symlink}" ||
+            exit 1
+        }
+      fi
+      actioned+=( ${plugin} )
+    fi
+  done
+
+  # Print what plugins where enabled/disabled if needed
```
StafDehat: Typo: s/where/were/

src/core/plugins
```diff
@@ -0,0 +1,306 @@
...
+
+      # Disable enabled plugin
+      elif [[ "${action}" == "disable" ]]; then
+        {
+          rm "${symlink}" ||
+            exit 1
+        }
+      fi
+      actioned+=( ${plugin} )
+    fi
```
StafDehat, line 215: This would be a good spot to log debug info about skipping a plugin, due to it already being in the desired state.

StafDehat (changes requested):

src/core/plugins
tonyskapunk: Your comment is correct.  I know this is an *assumption* but any plugin with spaces in its name is not expected to work, it has been documented in here: https://github.com/rackerlabs/recap/blob/master/README.md#naming-conventions

src/core/plugins
tonyskapunk: Ah, right, I'm getting rid of the `print_plugins` and re-using the above (newly renamed) function `get_plugin_by_status`, also moving the validation into a new function `validate_plugin_status`

tonyskapunk:
Thanks for all the great feedback @StafDehat !

Coming commit should cover most the feedback given.

Some additional changes were required as well, here a gist of the changes:

Renamed few functions:
- `csv_remove_dups()` --> `csv_to_list()`
- `plugin_validate_action()` --> `validate_plugin_action()`
- `get_plugin_status()` --> `get_plugin_by_status()`
- `validate_plugin()`--> `validate_plugin_name()`

Newly added:
- `validate_plugin_status()`
- `enable_plugin()`
- `disable_plugin()`

Deleted:
- `print_plugins()`

---

The only pending change is `enable_plugin_conf()` from your feedback: https://github.com/rackerlabs/recap/pull/224#discussion_r302747782 you're absolutely right, is overcomplicated, and yes, the end goal is to have a single config in it when enabling plugins.  The complexity comes when attempting to add the config in a particular place and avoiding adding additional entries.  I honestly like your suggestion a lot more, because is easy, let me think just any reasons why that not be good and then I'll submit that change as well.

tonyskapunk:
Including now your suggestion, thanks! 

I just modified a bit the query regex, since we don't care about those already commented out, i.e.

with:  `^\s*(#.*)?\s*USEPLUGINS\s*=`
```
$ echo '
> USEPLUGINS="no"
> USEPLUGINS="no" # comment
>   USEPLUGINS="no"
>   USEPLUGINS="no" #comment
> #comment   USEPLUGINS="no"
> # comment   USEPLUGINS="no"
>   # comment   USEPLUGINS="no"
> '  | grep -P '^\s*(#.*)?\s*USEPLUGINS\s*='
USEPLUGINS="no"
USEPLUGINS="no" # comment
  USEPLUGINS="no"
  USEPLUGINS="no" #comment
#comment   USEPLUGINS="no"
# comment   USEPLUGINS="no"
  # comment   USEPLUGINS="no"
``` 

with: `^\s*USEPLUGINS='
```
$ echo '
USEPLUGINS="no"
USEPLUGINS="no" # comment
  USEPLUGINS="no"
  USEPLUGINS="no" #comment
#comment   USEPLUGINS="no"
# comment   USEPLUGINS="no"
  # comment   USEPLUGINS="no"
'  | grep -P '^\s*USEPLUGINS='
USEPLUGINS="no"
USEPLUGINS="no" # comment
  USEPLUGINS="no"
  USEPLUGINS="no" #comment
```

tonyskapunk:
I'll go ahead and merge this one, I've implemented all the suggested changes and is doing what's expected. Can always revisit prior merging into master.

Issue #228, opened 2019-11-11
Non-exec scripts


tonyskapunk:
When building the recap rpm, rpmlint reports multiple files with a `non-executable-script` error. Let's get that fixed.

```
recap.noarch: E: non-executable-script /usr/lib/recap/core/fdisk 644 /bin/bash 
recap.noarch: E: non-executable-script /usr/lib/recap/core/mysql 644 /bin/bash 
recap.noarch: E: non-executable-script /usr/lib/recap/core/netstat 644 /bin/bash 
recap.noarch: E: non-executable-script /usr/lib/recap/core/ps 644 /bin/bash 
recap.noarch: E: non-executable-script /usr/lib/recap/core/pstree 644 /bin/bash 
recap.noarch: E: non-executable-script /usr/lib/recap/core/resources 644 /bin/bash 
recap.noarch: E: non-executable-script /usr/lib/recap/core/send_mail 644 /bin/bash 
recap.noarch: E: non-executable-script /usr/lib/recap/plugins-available/docker_top 644 /bin/bash 
recap.noarch: E: non-executable-script /usr/lib/recap/plugins-available/http_status 644 /bin/bash 
recap.noarch: E: non-executable-script /usr/lib/recap/plugins-available/kernel_cmd 644 /bin/bash 
recap.noarch: E: non-executable-script /usr/lib/recap/plugins-available/redis 644 /bin/bash 
recap.noarch: E: non-executable-script /usr/lib/recap/plugins-available/system_locks 644 /bin/bash 
```

Source: https://taskotron.fedoraproject.org/artifacts/all/b2473242-dfa3-11e9-b225-52540077ca13/tests.yml/recap-2.1.0-3.fc30.log

The solution is super easy, from http://wiki.rosalab.ru/en/index.php/Rpmlint_Errors

> **non-executable-script**
> 
> This text file contains a shebang or is located in a path dedicated for executables, but lacks the executable bits and cannot thus be executed. If the file is meant to be an executable script, add the executable bits, otherwise remove the shebang or move the file elsewhere.

PR #229, opened 2019-11-11
Removed shebang per issue #228

stephenlang:


tonyskapunk (approved):
This is amazing, thanks for taking care of this so quickly @stephenlang !!

Just one minor note, usually in the description above you will add human friendly details of your change, e.g.

> Removing shebangs on files that are sourced by `recap`
> Fixes #228 

Other than that everything looks right Thanks!

PR #230, opened 2019-11-11
Add el8 to CI

tonyskapunk:
CentOS 8 was released not long ago, and it was missing from the CI environment, adding it.


tonyskapunk:
`elinks` is not available in the default repos on CentOS 8, need to tweak the dependency installer to enable the right repo.

https://travis-ci.org/rackerlabs/recap/jobs/610593621#L158-L162

PR #233, opened 2020-07-18
Add USEMYSQLFULLPROCESSLIST option to mysql plugin

### LINKED ISSUE #231: option: mysql show processlist instead of show full processlist
It would be nice to have a non-FULL processlist configuration that continues to default to full processlist (call a new print_mysql_full_procs):

> USEMYSQLFULLPROCESSLIST=yes

> print_mysql_procs()
> ..
>
> mysqladmin processlist

> print_mysql_full_procs()
> mysqladmin -v processlist

Whether the above is implemented or not, it would be good to correct existing logging which incorrectly claims that "show processlist" was issued.

This feature request is to avoid having to suppress all error messages going to email from cron, when you regularly hit an error around max_allowed_packet, as might happen if you are often doing a large INSERT for bulk loading (eg you do not want to change the defaults for all local clients)

tonyskapunk:
Fixes #231 

- Modifies the `USEMYSQLPROCESSLIST` to be truly "show processlist"
- Adds a new option `USEMYSQLFULLPROCESSLIST`, this option perfoms a "show full processlist", this was the original functionality of `USEMYSQLPROCESSLIST`

---

Ran a test that shows the two different outputs ("show processlist" and "show full processlist") in this [gist](https://gist.github.com/tonyskapunk/2429607c0e21c9a09f35e887c6b5e7da)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased] -
 - Modify iotop default [#218][218]
 - Add plugin cli support to `recap` [#217][217]
+- Add USEMYSQLFULLPROCESSLIST option to mysql plugin [#231][231]
 
 ## [2.1.0] - 2019-04-12
 - Remove signal `17` in `recaplog` [(#208)][208]
--- a/src/core/mysql
+++ b/src/core/mysql
@@ -94,11 +94,18 @@ print_mysql_innodb_status() {
   log INFO "Ended 'mysql innodb' report"
 }
 
-# Print the output of "mysqladmin processlist" to the mysql file
+# Print the output of "mysqladmin [full] processlist" to the mysql file
 print_mysql_procs() {
-  log INFO "Starting 'mysql processlist' report - ${LOGFILE##*/}"
   local LOGFILE="$1"
   local MYCNF="$2"
+  local FULL="$3"
+  local MY_V=""
+  local MY_PL="processlist"
+  if [[ -n "${FULL}" ]]; then
+    MY_PL="full processlist"
+    MY_V="-v"
+  fi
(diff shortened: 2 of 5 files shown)


mikegriffin:
Hello, with recent activity in another bug I was wondering if you would still consider merging this PR

tonyskapunk:
Hi @mikegriffin, It Is good to go, but I really would like to have someone else review it before merging, but it has been out for longer than a year; I'm not sure if anyone would have some free cycles to look at it :crying_cat_face:

mikegriffin:
Reviewing the patch, it appears to work on my system, with some minor problems:

* Code comment in `src/core/mysql` should probably say "mysql [full] processlist"

* When all MySQL options are set to yes, you get both processlist and full processlist output. I suspect we want something like the following in `/usr/sbin/recap`:

```
375,378d374
<     if [[ "${USEMYSQLPROCESSLIST,,}" == "yes" ]]; then
<       print_blankline "${ITEM_FILE}"
<       print_mysql_procs "${ITEM_FILE}" "${MYCNF}"
<     fi
381a378,382
>       else
>               if [[ "${USEMYSQLPROCESSLIST,,}" == "yes" ]]; then
>               print_blankline "${ITEM_FILE}"
>               print_mysql_procs "${ITEM_FILE}" "${MYCNF}"
>               fi
```

tonyskapunk:
:wave:  @mikegriffin sorry for the lack of response here and thanks for the review.

> Code comment in src/core/mysql should probably say "mysql [full] processlist"

This is fixed in my next commit

> When all MySQL options are set to yes, you get both `processlist` and `full processlist` output

In that case, I'm making `full processlist` to get priority over `processlist`. If both settings are enabled.

Please give it a look, and if everything looks good I'll merge it

mikegriffin (approved):
LGTM

PR #237, opened 2020-11-20
Add el8 to CI

tonyskapunk:
- Adds el8 to CI
- Splits dependencies per distro


tonyskapunk:
:facepalm: Need to adjust this to be in dev, closing to fix it

tonyskapunk:
Well, it is not needed because is *already* in the `development` added in #230

Issue #238, opened 2020-12-24
MySQL InnoDB status report is not retrieved if MySQL tmpdir points to a symlink


tonodor:
I have found that if MySQL tmpdir points to a symlink, recap fails to retrieve the InnoDB status report.
This is because it compares the path of InnoDB temporary files with the tmpdir value but fails because the paths contain the destination folder:

```
# mysql -Bse "SHOW VARIABLES LIKE 'tmpdir';"
tmpdir  /var/lib/mysqltmp

# ll /var/lib/mysqltmp
lrwxrwxrwx. 1 root root 14 Dec 24 11:39 /var/lib/mysqltmp -> /data/mysqltmp

# ll /proc/7305/fd | grep mysqltmp
lrwx------. 1 mysql mysql 64 Dec 24 11:44 11 -> /data/mysqltmp/ibywyLPT (deleted)
lrwx------. 1 mysql mysql 64 Dec 24 11:44 4 -> /data/mysqltmp/ibdgvlds (deleted)
lrwx------. 1 mysql mysql 64 Dec 24 11:44 5 -> /data/mysqltmp/ib0yHSjl (deleted)
lrwx------. 1 mysql mysql 64 Dec 24 11:44 6 -> /data/mysqltmp/ibphpqqe (deleted)
lrwx------. 1 mysql mysql 64 Dec 24 11:44 7 -> /data/mysqltmp/ibntvYF0 (deleted)
```

I'm going to submit a PR to address this.

tonyskapunk:
Hi, @tonodor thanks for reporting this issue, and thanks for providing a PR. I'll review and provide feedback there.

tonyskapunk:
Fixed in #239

PR #239, opened 2020-12-24
Fixed InnoDB report not working when tmpdir is a symlink

tonodor:
Re #238
I just changed print_mysql_innodb_status to add a check if TMPDIR is a symlink and eventually replace replace it with the destination.


tonyskapunk (approved):
The changes look good to me. The failures in the build were unrelated to this change.

Thanks :tada: !!

Issue #240, opened 2020-12-29
CentOS 8 Image from docker.io failing on build


tonyskapunk:
https://travis-ci.org/github/rackerlabs/recap/jobs/752021765#L147

Found that the image used by Travis CI is failing as PowerTools repo may not be available.

```
$ podman images|grep -P "centos\s*8"
docker.io/library/centos                          8       300e315adb2f  3 weeks ago    217 MB
registry.centos.org/centos                        8       c84cb20b344d  6 weeks ago    222 MB
```

The image from library/centos is newer and failing, the one from registry.centos.org is working as expected:

- library/centos:

  ```
  $ podman exec recap_on_${DISTRO/:/_} /recap/tests/install_deps.sh ${DISTRO}
  Error: Unknown repo: 'PowerTools'
  ```

- registry.centos.org:

  ```
  $ podman exec recap_on_${DISTRO/:/_} /recap/tests/install_deps.sh ${DISTRO}CentOS-8 - AppStream                            5.3 MB/s | 6.3 MB     00:01    
  CentOS-8 - Base                                 2.1 MB/s | 2.3 MB     00:01    
  CentOS-8 - Extras                                11 kB/s | 8.6 kB     00:00    
  CentOS-8 - PowerTools                           1.5 MB/s | 2.0 MB     00:01    
  Package iproute-5.3.0-1.el8.x86_64 is already installed.
  Dependencies resolved.
  ================================================================================
   Package          Arch    Version                             Repository   Size
  ================================================================================
  Installing:
   elinks           x86_64  0.12-0.58.pre6.el8                  PowerTools  916 k
   iotop            noarch  0.6-16.el8                          BaseOS       66 k
   make             x86_64  1:4.2.1-10.el8                      BaseOS      498 k
   psmisc           x86_64  23.1-5.el8                          BaseOS      151 k
   sysstat          x86_64  11.7.3-5.el8                        AppStream   425 k
  Upgrading:
   iproute          x86_64  5.3.0-5.el8                         BaseOS      665 k
  Installing dependencies:
   gpm-libs         x86_64  1.20.7-15.el8                       AppStream    39 k
   lm_sensors-libs  x86_64  3.4.0-21.20180522git70f7e08.el8     BaseOS       59 k  
  Transaction Summary
  ================================================================================
  Install  7 Packages
  Upgrade  1 Package
  ```

tonyskapunk:
The package that provides the `PowerTools` repository changes from 8.2 to 8.3:

[...]

PR #241, opened 2020-12-30
Update powertool repo name

tonyskapunk:
Fix: #240 

- Update powertool repo name


tonyskapunk:
It worked fine this time, https://travis-ci.org/github/rackerlabs/recap/builds/752080769 the failure is in the centos 6 build, but since it is EOL I'll retire that and any other EOL OS in another issue.

Issue #242, opened 2020-12-30
Retire EOL distros from build


tonyskapunk:
While working on #240 I found some EOL distros are failing, at this point CentOS 6 and Debian 6 are EOL

- CentOS 6 EOL [2020-11-20](https://wiki.centos.org/About/Product)
- Debian 8 (jessie) [2020-06-30](https://www.debian.org/News/2020/20200709)

PR #243, opened 2020-12-30
Remove EOL distros from build

tonyskapunk:
Fix: #242 

- Removes CentOS 6 and Debian 8 from builds


tonyskapunk:
Finally, all the checks have been passed.

